PR #7, opened 2018-01-10
Additional changes, trying to stay in sync before EOS

jtwill98:


cdearborn:
Hey JT, src/pilot/job.patch should also be removed in this PR.

gaelrehault:
@jtwill98 , few more files need to be removed : 
src/deploy/setup/setup_usb_idrac.log
src/pilot/constants.patch
src/pilot/node.patch
src/pilot/wsman.patch

PR #8, opened 2018-02-07
CES-9172: introspection fails due to locked node

cdearborn:
This patch changes the introspection logic so that it waits for OOB
introspection to complete before attempting to change the node state to
available.


rperryman:
LGTM - but I am not a coder...

PR #11, opened 2018-02-09
Patch support for NVMe drives into python-dracclient

cdearborn:
This patch adds support for NVMe drives to python-dracclient. A patch has been submitted upstream for this work, so we will need to remove this downstream patch eventually.

See: https://review.openstack.org/#/c/542911/


jtwill98:
+1, I'm good with these changes

PR #12, opened 2018-02-13
profile support

gaelrehault:


src/deploy/osp_deployer/profile.py
```diff
@@ -0,0 +1,120 @@
...
+                                  "::" + set + \
+                                  " is not valid for the profile " + \
+                                  self.profile_name + \
+                                  " it should be to one of the following " + \
+                                  "options : " + \
+                                  json.dumps(allowed_settings)
+                    elif validated is True:
+                        pass
+                    else:
+                        usr_sett = != user_config.get(str(stanza), set)
```
cdearborn: Looks like a syntax error on the above line? "= !="

src/deploy/osp_deployer/settings_sanity.py
```diff
@@ -50,7 +51,7 @@ def check_files(self):
         logger.debug("Check settings ip's are valid.")
         shouldbbevalidips = [
             'public_api_gateway',
-            'public_api_netmask', 'public_api_allocation_pool_start',
+            'public_api_netmask',
```
cdearborn: We still want public_api_allocation_pool_start in here don't we?

src/deploy/osp_deployer/settings/sample-fx.properties
```diff
@@ -1,197 +0,0 @@
-[
```
cdearborn: Would you mind moving the removal of this file into the PR for removing FX support?

src/deploy/osp_deployer/settings_sanity.py
gaelrehault: That particular validation moved into the profiles json file so no longer required here.

src/deploy/osp_deployer/settings/sample-fx.properties
gaelrehault: done

src/deploy/osp_deployer/profile.py
gaelrehault: thanks, fixed

jtwill98:
+1

PR #17, opened 2018-02-24
NIC configuration for 25G NICs

cdearborn:
This patch adds the NIC configuration files for 25G NICs.  It also changes the
NIC config infrastructure so that new NIC configurations can be added without
making any code changes.

### DIFF
--- a/src/deploy/osp_deployer/settings_sanity.py
+++ b/src/deploy/osp_deployer/settings_sanity.py
@@ -19,8 +19,8 @@ from settings.config import Settings
 from profile import Profile
 import logging
 import os.path
-import socket
 import collections
+import yaml
 
 logger = logging.getLogger("osp_deployer")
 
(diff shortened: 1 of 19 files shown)


src/deploy/osp_deployer/settings_sanity.py
```diff
@@ -489,3 +489,54 @@ def verify_iha_dependency_on_fencing(self):
...
+            missing_settings.append('nic_env_file')
+
+        if 'sah_bond_opts' not in ini_nics_settings:
+            missing_settings.append('sah_bond_opts')
+
+        extra_settings = set()
+        missing_variable_settings = set()
+        if not nic_env_file_missing:
+            # Get the NIC settings from the NIC environment file
+            with open(self.settings.nic_env_file_path, 'r') as yaml_stream:
```
gaelrehault: Maybe check if the file exists before trying to open it; not an issue for XSP and CSP profiles , but could potentially be for a custom profile/custom nic file

rperryman:
Successful deploy using new 4-nic configuration.  Sanity Test passes.

src/deploy/osp_deployer/settings_sanity.py
```diff
@@ -489,3 +489,54 @@ def verify_iha_dependency_on_fencing(self):
...
+        if 'nic_env_file' not in ini_nics_settings:
+            nic_env_file_missing = True
+            missing_settings.append('nic_env_file')
+
+        if 'sah_bond_opts' not in ini_nics_settings:
+            missing_settings.append('sah_bond_opts')
+
+        extra_settings = set()
+        missing_variable_settings = set()
+        if not nic_env_file_missing:
```
gaelrehault: if nic_env_file is missing from the ini file - it will return an error with only that setting missing logged (even tho you might have additional ones missing)
Would be best to also check the rest of the settings so all are reported in one shot.

cdearborn: It will report at least on the status of both nic_env_file and sah_bond_opts if one or the other or both are missing.
Because the remaining NIC configuration parameters will be different depending on the NIC configuration files, we have to validate the remaining NIC configuration settings in the user's .ini file with the NIC configuration settings in the nic_env_file.  As a result, if nic_env_file is missing then we can't validate the remaining NIC configuration settings.

src/deploy/osp_deployer/settings_sanity.py
cdearborn: Done.

PR #18, opened 2018-03-01
added test to abort the director install if customized overcloud... 

jtwill98:
…image fails


src/pilot/install-director.sh
```diff
@@ -193,6 +193,11 @@ echo
 if [ -n "${overcloud_nodes_pwd}" ]; then
     echo "# Setting overcloud nodes password"
     virt-customize -a overcloud-full.qcow2 --root-password password:$overcloud_nodes_pwd
+    if [ $? -ne 0 ]; then
```
gaelrehault: If you specified an overcloud password to be set, this check will pass even tho the customize image script might have failed.

PR #19, opened 2018-03-06
CES-9369: Remove sample fx files for deployment

cdearborn:
This commit removes two fx related files that we missed removing on the first
pass.


heydog15:
LGTM

PR #20, opened 2018-03-07
CES-9330 - Fixed formatting in network overlap error messages

gaelrehault:


gaelrehault:
note  : checks/build is passing, should not have used "error" in commit message  ...

src/deploy/osp_deployer/settings_sanity.py
```diff
@@ -507,11 +507,6 @@ def validate_nic_configs(self):
         extra_settings = set()
         missing_variable_settings = set()
         if not nic_env_file_missing:
-            if not os.path.isfile(self.settings.nic_env_file_path):
```
cdearborn: It looks like this piece of code was accidentally removed.

gaelrehault:
updated, thanks @cdearborn

PR #23, opened 2018-03-08
Removing rhscon and create new dashboard vm 

jtwill98:


gaelrehault:
src/pilot/config_dashboard.py still has a bunch of rhscon references, assuming this will be addressed later ? 

Also references in : 
src/update_upgrade/ceph_upgrade_director.sh
src/update_upgrade/ceph_upgrade_sah.sh
src/update_upgrade/upgrade-js6.0.1.sh

We need to talk about how we want to handle those with @dpaterson

heydog15:
In src/pilot/config_dashboard.py, the references to the BZ needs to be taken out.  Won't apply to Dashboard and that issue was fixed in ceph-ansible a while back.

Also in deploy-dashboard-vm.py, the Copyright needs to have 2018 added. Shouldn't the  dashboard-post.log be copied into the dashboard-ks-logs directory so all logs are together?

jtwill98:
I addressed some of the concerns.   Yes, I did not forget the update_upgrade scripts.  I will address these with Dave after fixing the config_dashboard.   What is here is the VM deployment and calls to it from automation.   The next step will be to finalize the changes config_dashboard script.  What's there current, is only the user and etc/hosts prep.  Running ansible to install the dashboard and collectors is part of the changes to finalize.

src/pilot/customize_image.sh
```diff
@@ -53,13 +53,12 @@ fi
 
 repos=(
     rhel-7-server-rpms
-    rhel-7-server-rhscon-2-agent-rpms
     rhel-7-server-rhceph-2-mon-rpms
 )
 
 packages=(
     calamari-server
```
dpaterson: We still need calamari server for cephmetrics?

jtwill98: That has yet to be determined.  I left it in during my investigation.  I will verify if it's needed during the integration effort.

src/deploy/osp_deployer/checkpoints.py
```diff
@@ -288,9 +288,9 @@ def rhscon_vm_health_check(self):
 
         logger.debug("*** Verify the Storage Console VM can ping "
```
pauldardeau: Nothing urgent, but we might need to change the name here away from Storage Console

src/deploy/osp_deployer/checkpoints.py
```diff
@@ -300,9 +300,9 @@ def rhscon_vm_health_check(self):
         logger.debug(
             "*** Verify the Storage Console VM can ping the SAH node "
```
pauldardeau: Same here

src/mgmt/dashboard.cfg
```diff
@@ -1,4 +1,4 @@
-# (c) 2014-2018 Dell
```
pauldardeau: Remove copyright indicator?

src/deploy/osp_deployer/checkpoints.py
jtwill98: NP.

cdearborn:
+1 Looks good.

PR #26, opened 2018-03-12
Numa/Huge pages support.

jproque-dell:


src/deploy/osp_deployer/director.py
```diff
@@ -865,6 +865,17 @@ def deploy_overcloud(self):
                                     " --ntp " + \
                                     self.settings.sah_node.provisioning_ip
 
+        if self.settings.hpg_enable is True:
+            cmd += " --enable_hugepage "
```
dpaterson: nit: should hugepage be plural hugepages?

src/pilot/templates/nic-configs-fx/dellcompute.yaml
```diff
@@ -0,0 +1,194 @@
+# Copyright (c) 2016 Dell Inc. or its subsidiaries.
```
rperryman: This files needs to be removed.

src/pilot/templates/nic-configs/dellcompute.yaml
```diff
@@ -0,0 +1,183 @@
+heat_template_version: 2015-04-30
```
rperryman: This file needs to be removed

src/pilot/deploy-overcloud.py
```diff
@@ -398,7 +1207,6 @@ def main():
               " --os-user-id {}" \
               " --os-password {}" \
               " --control-scale {}" \
-              " --compute-scale {}" \
```
gaelrehault: Is there no other way to do this and be more consistant and rather that edit the environment file pass in a parameter to reference the custom role here ? ( for manual deployments )

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,6 +171,29 @@ StorageBond1Interface1=em2
 StorageBond1Interface2=p2p2
 StorageBondInterfaceOptions=mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
 
+[Dell NFV Settings]
```
cdearborn: There is a second Dell NFV Settings section here:
https://github.com/dsp-jetpack/JetPack/pull/26/files#diff-03907ea6ba7811b5a8d4acb7b8c60701L255
One or the other needs to be removed.

src/pilot/customize_image.sh
```diff
@@ -87,5 +87,12 @@ virt-customize \
     --sm-unregister \
     --selinux-relabel 2>&1 | tee -a ~/pilot/customize_image.log
 
+virt-customize \
+	--add overcloud-full.qcow2 \
+	--run-command "cd /usr/share/openstack-puppet/modules;puppet module generate dell-dellnfv --skip-interview" \
+    --run-command "mv /usr/share/openstack-puppet/modules/dell-dellnfv /usr/share/openstack-puppet/modules/dellnfv" \
+    --link /usr/share/openstack-puppet/modules/dellnfv:/etc/puppet/modules/dellnfv \
```
cdearborn: Can you fix the indenting on the above 2 lines?

src/pilot/deploy-overcloud.py
```diff
@@ -34,6 +39,767 @@
...
+class ConfigOvercloud(object):
+    """
+    Description: Class responsible for overcloud configurations.
+    """
+
+    def __init__(self, nova):
+        self.nova = nova
+        self.aggregate_obj = aggregates.AggregateManager(self.nova)
+
+    @classmethod
```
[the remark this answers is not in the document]
cdearborn: I agree with this.  CredentialHelper.get_overcloud_creds() should be used, and that would eliminate this method.

src/pilot/deploy-overcloud.py
```diff
@@ -34,6 +39,767 @@
...
+        except Exception as error:
+            message = "Exception {}: {}".\
+                      format(type(error).__name__, str(error))
+            print "{}".format(message)
+            raise Exception("Failed to get overcloud details \
+                            from the overcloud rc file {}".format(file_path))
+
+    @classmethod
+    def get_undercloud_details(cls):
+        try:
```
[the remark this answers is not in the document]
cdearborn: I agree again. CredentialHelper.get_undercloud_creds() should be used instead, which would eliminate this method.

src/pilot/deploy-overcloud.py
```diff
@@ -34,6 +39,767 @@
...
+    """
+
+    def __init__(self, nova):
+        self.nova = nova
+        self.aggregate_obj = aggregates.AggregateManager(self.nova)
+
+    @classmethod
+    def get_overcloud_details(cls, overcloud_name):
+        try:
+            print "Getting overcloud details"
```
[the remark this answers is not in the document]
cdearborn: definitely should use logger statements instead of print statements throughout this file.

src/pilot/templates/dell-environment.yaml
```diff
@@ -44,6 +55,13 @@ parameter_defaults:
 
   # Additional hiera data for all nodes
   ExtraConfig:
+    # NFV Profile settings start here
+    dellnfv::hugepages::enable: True
+    dellnfv::hugepages::hugepagesize: 2M
+    dellnfv::hugepages::hugepagecount: 49152
+    dellnfv::numa::enable: True
+    dellnfv::numa::vcpu_pin_set: \"8-47\"
+
```
gaelrehault: The above dellnfv::hugepages::enable & dellnfv::numa::enable should be set to False since XSP is our default profile.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
gaelrehault: This also needs to go in the sample_xsp_profile.ini ( with the values disabled accordingly)

gaelrehault:
Could we make the src/pilot/templates/DellNfv  directory lowercase ( dellnfv or dell_nfv )

src/deploy/osp_deployer/deployer.py
```diff
@@ -231,7 +231,14 @@ def deploy():
                                     settings.director_install_account_pwd,
                                     cmd)[0]
         logger.debug("=== Overcloud stack state : " + overcloud_status)
-
+        if settings.hpg_enable:
```
gaelrehault: There is a retreive_nodes_ips method in director.py logging the  deployment related information , those should probably go in there.

src/deploy/osp_deployer/settings/config.py
```diff
@@ -239,6 +239,16 @@ def __init__(self, settings_file):
         else:
             self.enable_instance_ha = False
         self.overcloud_nodes_pwd = deploy_settings['overcloud_nodes_pwd']
+        dellnfv_settings = self.get_settings_section(
+            "Dell NFV Settings")
+        if dellnfv_settings['hpg_enable'].lower() == 'true':
+            self.hpg_enable = True
```
gaelrehault: add and else , set to False

src/deploy/osp_deployer/settings/config.py
```diff
@@ -239,6 +239,16 @@ def __init__(self, settings_file):
...
             self.enable_instance_ha = False
         self.overcloud_nodes_pwd = deploy_settings['overcloud_nodes_pwd']
+        dellnfv_settings = self.get_settings_section(
+            "Dell NFV Settings")
+        if dellnfv_settings['hpg_enable'].lower() == 'true':
+            self.hpg_enable = True
+        self.hpg_flavor_name_list = dellnfv_settings['hpg_flavor_name_list']
+        self.hpg_size = dellnfv_settings['hpg_size']
+        if dellnfv_settings['numa_enable'].lower() == 'true':
+            self.numa_enable = True
```
gaelrehault: add and else , set to False

src/pilot/assign_role.py
```diff
@@ -93,7 +93,7 @@
 ROLES = {
     'controller': 'control',
     'compute': 'compute',
-    'storage': 'ceph-storage'
+    'storage': 'ceph-storage',
```
gaelrehault: remove last comma on last list item

src/pilot/deploy-overcloud.py
```diff
@@ -328,6 +1127,12 @@ def main():
             raise ValueError("\nError: {} failed, unable to continue.  See "
                              "the comments in that file for additional "
                              "information".format(cmd))
+        # Pass the parameters required by puppet which will be used
+        # to enable/disable dell nfv features
+        # Edit the dellnfv_environment.yaml
+        edit_dell_environment_file(args.enable_hugepage, args.enable_numa,
+                                   args.hugepage_size, vcpu_pin_set,
```
gaelrehault: i suspect this will fail if those params are not passed  as args, since they dont have a default value

src/deploy/osp_deployer/deployer.py
cdearborn: I think that's a fine idea, but let's change the name of the method to something like "summarize_deploment" or something like that.  I displays information about the overcloud nodes, but also displays endpoints too.

src/pilot/templates/DellNfv/hugepages.pp
```diff
@@ -0,0 +1,40 @@
+#class: dellnfv::hugepages
```
cdearborn: Should add the standard licensing header to the top of this file.

src/pilot/templates/DellNfv/hugepages.yaml
```diff
@@ -0,0 +1,29 @@
+heat_template_version: 2016-04-08
```
cdearborn: Should add the standard licensing header to the top of this file.

src/pilot/templates/DellNfv/numa.pp
```diff
@@ -0,0 +1,47 @@
+# == Class: dellnfv::numa
```
cdearborn: Should add the standard licensing header to the top of this file.

src/pilot/templates/DellNfv/numa.yaml
```diff
@@ -0,0 +1,29 @@
+heat_template_version: 2016-04-08
```
cdearborn: Should add the standard licensing header to the top of this file.

src/pilot/templates/dell-environment.yaml
```diff
@@ -88,7 +104,7 @@ parameter_defaults:
         journal: '/dev/sdc'
       '/dev/sdo':
         journal: '/dev/sdc'
-
+
```
cdearborn: Can you remove the extra spaces on the line above?

src/pilot/templates/network-isolation.yaml
```diff
@@ -0,0 +1,70 @@
+# Enable the creation of Neutron networks for isolated Overcloud
```
cdearborn: Should put the standard licensing head at the top of this file.

src/pilot/templates/roles_data.yaml
```diff
@@ -0,0 +1,200 @@
+# Specifies which roles (groups of nodes) will be deployed
```
cdearborn: Should put the standard license header at the top of this file.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
 
+logging.basicConfig()
+logger = logging.getLogger(os.path.splitext(os.path.basename(sys.argv[0]))[0])
+
 home_dir = os.path.expanduser('~')
 
 BAREMETAL_FLAVOR = "baremetal"
 
+# Dell nfv configuration global values
+undercloudrc_file_name = "stackrc"
```
cdearborn: I think the above variable may no longer be used.  Can you check and remove if this is the case?

src/pilot/deploy-overcloud.py
```diff
@@ -430,19 +997,27 @@ def main():
...
         end = time.time()
-        print '\nExecution time: {} (hh:mm:ss)'.format(
-            time.strftime('%H:%M:%S', time.gmtime(end - start)))
+        logger.info('\nExecution time: {} (hh:mm:ss)'.format(
+            time.strftime('%H:%M:%S', time.gmtime(end - start))))
         print 'Fetching SSH keys...'
         update_ssh_config()
         if status == 0:
             horizon_url = finalize_overcloud()
+            print("\nDeployment Completed")
```
cdearborn: Need to convert the above print to a logging statement or remove it.

src/pilot/deploy-overcloud.py
```diff
@@ -342,9 +908,14 @@ def main():
...
 
+        # The roles_data.yaml must be included at the beginning.
+        # This is needed to enable the custome role Dell Compute.
+        # It overrides the default roles_data.yaml
+        env_opts = "-r ~/pilot/templates/roles_data.yaml"
+
         # The network-environment.yaml must be included after the
         # network-isolation.yaml
-        env_opts = "-e ~/pilot/templates/overcloud/environments/" \
+        env_opts = " -e ~/pilot/templates/overcloud/environments/" \
```
cdearborn: The above change isn't necessary.  Can you revert it?

cdearborn:
The variable compute_flavor in deploy-overcloud.py is no longer used and should be removed.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+                # Add hosts to aggregate
+                self.add_hosts_to_aggregate(aggregate_id, host_name_list)
+        except Exception as error:
+            message = "Exception {}: {}".\
+                      format(type(error).__name__, str(error))
+            logger.error("{}".format(message))
+            raise Exception("Failed to configure \
+                            dell aggregate {}.".format(aggregate_name))
+
+    def create_flavor(self, flavor_name):
```
cdearborn: We already have a "create_flavors" method that does all of the work below.  We really shouldn't have 2 different pieces of code for doing the same thing.  I think i would modify this method so that it accepts memory, disk, cpus, and id, then modify the create_flavors method to call into this one.  Probably should rename create_flavors to create_default_flavors or something like that.

src/pilot/deploy-overcloud.py
```diff
@@ -34,6 +39,767 @@
...
+
+class NodeConfig:
+    """
+    Description: Class responsible for node configuration operations
+    """
+
+    def __init__(self, ip_address):
+        self.ip_address = ip_address
+
+    def ssh_connect(self):
```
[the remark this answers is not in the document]
cdearborn: I agree, we really should use command_helper.Ssh.get_client()

src/pilot/deploy-overcloud.py
```diff
@@ -34,6 +39,767 @@
...
+        except Exception as error:
+            message = "Exception {}: {}". \
+                      format(type(error).__name__, str(error))
+            print "Failed to establish SSH connection to remote machine {} " \
+                  "with error {}".format(self.ip_address, message)
+
+    def close_ssh_connection(self):
+        self.ssh_connection.close()
+
+    def execute_cmd(self, command):
```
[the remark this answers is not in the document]
cdearborn: I agree, we really should use command_helper.Ssh.execute_command()

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+              'auth_url': os_auth_url,
+              'project_id': os_tenant_name}
+
+    # Create nova client object
+    nova = nova_client.Client(2, **kwargs)
+
+    # Create the ConfigOvercloud object
+    config_oc_obj = ConfigOvercloud(nova)
+    dell_compute_nodes_uuid = config_oc_obj.get_dell_compute_nodes_uuid()
+    logger.info("{}".format(dell_compute_nodes_uuid))
```
cdearborn: The above logging statement should probably be at the debug level, and should include some indication as to what is being logged like "dell_compute_nodes_uuid: {}".format ...

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+
+
+def create_custom_flavors(overcloud_name,
+                          enable_hugepage,
+                          enable_numa, hugepage_size,
+                          hpg_flavor_names_list,
+                          numa_flavor_names_list):
+    try:
+        if enable_hugepage or enable_numa:
+            logger.info("Create custom flavors")
```
cdearborn: "Creating custom flavors" above

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+                      'password': os_password,
+                      'auth_url': os_auth_url,
+                      'project_id': os_tenant_name}
+
+            # Create nova client object
+            nova = nova_client.Client(2, **kwargs)
+
+            # Create the ConfigOvercloud object
+            config_oc_obj = ConfigOvercloud(nova)
+            # Create falvors for HugePage feature and set its metadata
```
cdearborn: "falvors" should be "flavors" above

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+            if enable_hugepage:
+                metadata = ConfigOvercloud.get_hp_flavor_meta(hugepage_size)
+                for flavor_name in hpg_flavor_names_list:
+                    flavor = config_oc_obj.create_flavor(flavor_name)
+                    config_oc_obj.set_flavor_metadata(flavor, metadata)
+
+            # Create falvors for NUMA feature and set its metadata
```
cdearborn: "falvors" should be "flavors" above

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+        finally:
+            # Close the ssh connection
+            self.ssh_connection.close()
+
+
+def edit_dell_environment_file(enable_hugepage,
+                               enable_numa,
+                               hugepage_size,
+                               vcpu_pin_set,
+                               dell_compute_count=0):
```
cdearborn: Can you change the above line to not default to anything so that dell_compute_count is required?  It looks like it is only called from 1 place, and it passes the compute count.  A value of 0 for the compute count really is invalid.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+                str(enable_hugepage) +
+                '|" ' +
+                file_path,
+                'sed -i "s|dellnfv::numa::enable:.*|' +
+                'dellnfv::numa::enable: ' +
+                str(enable_numa) +
+                '|" ' +
+                file_path]
+
+        if dell_compute_count > 0:
```
cdearborn: I don't think we need the above line since there should always be at least 1 compute

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+        if enable_numa:
+            cmds.append('sed -i "s|dellnfv::numa::vcpu_pin_set:.*|' +
+                        'dellnfv::numa::vcpu_pin_set: \\\\\\\\\\"' +
+                        vcpu_pin_set +
+                        '\\\\\\\\\\"|" ' +
+                        file_path)
+
+        for cmd in cmds:
+            status = os.system(cmd)
+            logger.info("cmd: {}".format(cmd))
```
cdearborn: The above logging line should be at the debug level.

src/pilot/deploy-overcloud.py
```diff
@@ -430,19 +997,27 @@ def main():
...
-        print '\nExecution time: {} (hh:mm:ss)'.format(
-            time.strftime('%H:%M:%S', time.gmtime(end - start)))
+        logger.info('\nExecution time: {} (hh:mm:ss)'.format(
+            time.strftime('%H:%M:%S', time.gmtime(end - start))))
         print 'Fetching SSH keys...'
         update_ssh_config()
         if status == 0:
             horizon_url = finalize_overcloud()
+            print("\nDeployment Completed")
+            create_aggregates(args.enable_hugepage, args.enable_numa,
```
cdearborn: It seems like this chunk of code (create_aggregates, create_custom_flavors, and update_filters should be moved into finalize_overcloud() since that method is responsible for doing these types of things.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+
+    def execute_cmd(self, command):
+        command = "sudo " + command
+        logger.info("Executing command {} on {}".format(command, self.ip_address))
+        time.sleep(1)
+        ssh_stdin, ssh_stdout, ssh_stderr = self.ssh_connection.cmd(command)
+        time.sleep(1)
+        return ssh_stdout.readlines(), ssh_stderr.readlines()
+
+    def update_filter(self):
```
cdearborn: I would really prefer you change this to use the method of setting NovaSchedulerDefaultFilters in dell-environment.yaml.  As you point out in your writeup, it's simpler with fewer moving parts.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,534 @@
...
+                        hugepage_size[0:-1] +
+                        '|" ' +
+                        file_path)
+            cmds.append('sed -i "s|dellnfv::hugepages::hugepagecount:.*|' +
+                        'dellnfv::hugepages::hugepagecount: ' +
+                        str(hugepage_number) +
+                        '|" ' +
+                        file_path)
+        if enable_numa:
+            cmds.append('sed -i "s|dellnfv::numa::vcpu_pin_set:.*|' +
```
cdearborn: Are you planning on changing this to use the technique that you wrote up? NovaVcpuPinSet

src/pilot/deploy-overcloud.py
jproque-dell: Yeah, it's planned already. However, it's likely to happen after this PR.

src/pilot/deploy-overcloud.py
jproque-dell: Agree. Removed, compute count  is at least one. It was needed at a time where we supposed it would have been possible to mix different type of compute (eg: regular + NFV ), meaning that our custom node count can be null.

src/pilot/deploy-overcloud.py
jproque-dell: Agree. Same comment as above. Removed.

src/deploy/osp_deployer/deployer.py
gaelrehault: @jproque-dell  look at director.py retreive_nodes_ips , that function logs all the deployment related informations ( nodes ips, urls, passwords etc.. ) both to the output and to a separate log file. Those nfv related info should go in there as well, like Chris is suggesting, just rename that method to something more generic.

src/pilot/deploy-overcloud.py
jproque-dell: Yeah, it's planned but it's likely to happen after the PR.

src/pilot/deploy-overcloud.py
jproque-dell: Changed to debug.

src/pilot/deploy-overcloud.py
jproque-dell: Changed.

src/pilot/deploy-overcloud.py
jproque-dell: Fixed.

src/pilot/deploy-overcloud.py
jproque-dell: Fixed.

src/pilot/deploy-overcloud.py
jproque-dell: Yeah, makes sense. Changed.

src/pilot/deploy-overcloud.py
```diff
@@ -256,6 +1030,22 @@ def main():
...
+                            required=False,
+                            help="HostOS Cpus")
+        parser.add_argument("--hugepage_flavor_list",
+                            dest="hugepage_flavor_list",
+                            required=False,
+                            help="Hugepage Flavor list ")
+        parser.add_argument("--numa_flavor_list",
+                            dest="numa_flavor_list",
+                            required=False,
+                            help="NUMA Flavor list")
```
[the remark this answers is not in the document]
gaelrehault: I m saying that if enable_numa & enable_hugepage are passed as True , then you should check that the  associated parameters (numa_flavor_list, hugepage_flavor_list, hostos_cpus, hugepage_size)  are also passed / not null. Prevents errors if a manual deployment is done ommmiting those parameters.

src/pilot/templates/DellNfv/hugepages.pp
jproque-dell: Added.

src/pilot/templates/DellNfv/hugepages.yaml
jproque-dell: Added.

src/pilot/templates/DellNfv/numa.pp
jproque-dell: Added.

src/pilot/templates/DellNfv/numa.yaml
jproque-dell: Added.

src/pilot/templates/network-isolation.yaml
jproque-dell: Done.

src/pilot/templates/roles_data.yaml
jproque-dell: Done.

src/deploy/osp_deployer/deployer.py
jproque-dell: Renamed retreive_node_ips to summarize_deployment as suggested by Chris and included all NFV Features info coming from deployer into this method.

src/pilot/deploy-overcloud.py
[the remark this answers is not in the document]
jproque-dell: Done.

src/pilot/deploy-overcloud.py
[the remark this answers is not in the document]
jproque-dell: All ssh methods are removed from this file. Updating filters and modifying nova configuration file are now handled by TripleO params: NovaSchedulerDefaultFilters and NovaCpuPinSet.

src/pilot/deploy-overcloud.py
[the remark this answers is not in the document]
jproque-dell: Same. All ssh methods are removed from this file. Updating filters and modifying nova configuration file are now handled by TripleO params: NovaSchedulerDefaultFilters and NovaCpuPinSet.

src/pilot/deploy-overcloud.py
jproque-dell: Agree. Moved.

src/pilot/deploy-overcloud.py
jproque-dell: Upon reflection and after some checks, it turns out it was easier to switch to the the TripleO params method. Update_filter is now useless, hence removed. Edit_dell_environment_file is now handling those params.

src/pilot/deploy-overcloud.py
jproque-dell: Upon reflection and after some checks, it turns out it was easier to switch to the the TripleO params method. Update_filter is now useless, hence removed. Edit_dell_environment_file is now handling those params.

src/pilot/deploy-overcloud.py
jproque-dell: It seems fine to me. Meanwhile, i would like to put this on hold, as there is a discussion on-going between you, gael and the rest of the team to decide the direction we'd like to take with respect to the flavors going forward. Besides, we need to set additional metadata to flavors used for CSP Profile.

gaelrehault:
hostos_cpus is referenced in the ini & passed in as a parameter/logged in several places, but i don't see it actually being used/set anywhere ?

gaelrehault:
As agreed, let s get rid of the flavours creation.

src/pilot/deploy-overcloud.py
```diff
@@ -256,6 +599,14 @@ def main():
...
                             default=None,
                             help="The name of the overcloud")
+        parser.add_argument("--hugepages_size",
+                            dest="hugepages_size",
+                            required=False,
+                            help="HugePages size")
+        parser.add_argument("--hostos_cpus",
+                            dest="hostos_cpus",
+                            required=False,
+                            help="HostOS Cpus")
```
gaelrehault: hostos_cpus arg . should be removed

jproque-dell: Agree. Removed.

src/deploy/osp_deployer/director.py
jproque-dell: Changed to plural.

src/deploy/osp_deployer/settings/config.py
```diff
@@ -239,6 +239,17 @@ def __init__(self, settings_file):
         else:
             self.enable_instance_ha = False
         self.overcloud_nodes_pwd = deploy_settings['overcloud_nodes_pwd']
+        dellnfv_settings = self.get_settings_section(
+            "Dell NFV Settings")
+        if dellnfv_settings['hpg_enable'].lower() == 'true':
+            self.hpg_enable = True
```
dpaterson: Should we have defaults, what happens if there is no Dell NFV Settings in ini?  So if  get_settings_section doesn't find a particular section there are a set of defaults returned?  I've seen this pattern in other openstack projects.  Perhaps this could be a future refactor.

gaelrehault: Making sure all stanza/options are present is already covered further up in that class

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +36,343 @@
...
+              'auth_url': os_auth_url,
+              'project_id': os_tenant_name}
+
+    n_client = nova_client.Client(2, **kwargs)
+    for compute in n_client.servers.list(detailed=False,
+                                         search_opts={'name': 'compute'}):
+        logger.info("Rebooting server with id: ".format(compute_name))
+        n_client.servers.reboot(compute.id)
+
+        # Wait for 5 mins to let dell nodes boot up
```
cdearborn: As written, this code will reboot a compute node and then sleep 5 minutes, repeat.  This means that the total sleep time will be number_of_compute_nodes*5 minutes.  I think what we want to do is reboot all the compute nodes, then sleep 5 minutes, so I think we just need to unindent the above line and the following line 4 spaces.

gaelrehault: Agreed - thread it up , and rather than sleep 5 mns , wait for the computes to come back up , you should find some examples in existing code how to do that.

cdearborn: Agreed that sleeping isn't the best way to do this.  A better way would be to wait until you can SSH into each of the computes.  You can do this with or without threads depending on how you want to do it.  I'm happy to leave this to a follow-on PR though.

jproque-dell: Added a ssh validation check to the reboot_compute_nodes method

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +37,370 @@
...
+
+    dell_compute = []
+    for compute in n_client.servers.list(detailed=False,
+                                         search_opts={'name': 'compute'}):
+        dell_compute.append(n_client.servers.ips(compute)['ctlplane'][0]['addr'])
+
+        logger.info("Rebooting server with id: ".format(compute_id))
+        n_client.servers.reboot(compute.id)
+
+        retries = 32
```
cdearborn: The way the code is written now, i t will reboot one server and then execute the following code to wait.  We want it to reboot all the servers and then execute the wait code.  You just need to unindent lines 370-394 by 4 spaces to make it work this way.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +37,370 @@
...
+                else:
+                    logger.info("Server with IP: {} is responsive".format(ip))
+                    ssh_success_count += 1
+                    break
+
+                # Waiting 10s before the next attempt
+                sleep(10)
+
+        if ssh_success_count == len(dell_compute):
+            logger.info("All compute nodes are now responsive. Continuing...")
```
cdearborn: Should add an "else" clause to the above "if" statement so that it logs an error and aborts if not all of the compute nodes come back up.

src/pilot/deploy-overcloud.py
jproque-dell: Well done! Fixed.

jproque-dell: Fixed.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +37,370 @@
...
+        logger.info("Rebooting server with id: ".format(compute_id))
+        n_client.servers.reboot(compute.id)
+
+        retries = 32
+        ssh_success_count = 0
+
+        # Wait for 30s for the nodes to be powered cycle then do the checks
+        sleep(30)
+
+        for ip in dell_compute:
```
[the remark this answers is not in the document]
jproque-dell: Fixed.

src/pilot/deploy-overcloud.py
jproque-dell: else statement added.

src/pilot/deploy-overcloud.py
```diff
@@ -430,19 +815,20 @@ def main():
         start = time.time()
         status = run_deploy_command(cmd)
         end = time.time()
-        print '\nExecution time: {} (hh:mm:ss)'.format(
-            time.strftime('%H:%M:%S', time.gmtime(end - start)))
+        logger.info('\nExecution time: {} (hh:mm:ss)'.format(
+            time.strftime('%H:%M:%S', time.gmtime(end - start))))
         print 'Fetching SSH keys...'
         update_ssh_config()
```
uday344: This method should only be called after successful deployment. so it should go under "if status == 0:" condition.

src/pilot/deploy-overcloud.py
```diff
@@ -430,19 +815,20 @@ def main():
...
         else:
             horizon_url = None
-        print 'Overcloud nodes:'
+        logger.info('Overcloud nodes:')
         identify_nodes()
 
         if horizon_url:
-            print '\nHorizon Dashboard URL: {}\n'.format(horizon_url)
+            logger.info('\nHorizon Dashboard URL: {}\n'.format(horizon_url))
     except ValueError as err:
```
cdearborn: The above line should be changed to:
except Exception as err:
so it picks up failure in the new create_* code and failure of compute nodes to reboot.

src/pilot/deploy-overcloud.py
```diff
@@ -30,18 +37,372 @@
...
+                ssh_success_count += 1
+                break
+
+            # Waiting 10s before the next attempt
+            sleep(10)
+
+    if ssh_success_count == len(dell_compute):
+        logger.info("All compute nodes are now responsive. Continuing...")
+    else:
+        logger.error("Failed to reboot the nodes or at least one node failed to get back up")
```
cdearborn: In addition to logging the error, we should raise an exception so the script aborts with an error code.

jproque-dell: Agree. Done.

src/pilot/deploy-overcloud.py
jproque-dell: Agree. Changed.

src/pilot/deploy-overcloud.py
```diff
@@ -30,14 +30,15 @@
...
+from logging_helper import LoggingHelper
 
 from credential_helper import CredentialHelper
 
 # Dell utilities
 from identify_nodes import main as identify_nodes
 from update_ssh_config import main as update_ssh_config
 
-logging.basicConfig()
+logging.basicConfig(level=logging.info)
```
gaelrehault: should be .INFO ( uppercase )

gaelrehault:
Merging. Thanks @jproque-dell

PR #27, opened 2018-03-14
removed left over fx yamls

gaelrehault:


heydog15:
LGTM.  Was going to write a jira ticket on this yesterday since I was in that directory and saw the fx stuff still there.

PR #28, opened 2018-03-15
OVS-DPDK support

jproque-dell:


src/deploy/osp_deployer/settings/sample_xsp_profile.ini
```diff
@@ -171,6 +171,30 @@ StorageBond1Interface1=em2
...
 StorageBondInterfaceOptions=mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
 
+[Dell NFV Settings]
+
+# OVS-DPDK Settings
+# ovs_dpdk_enable supports three values
+# ovs_dpdk_enable=None: disables OVS-DPDK
+# ovs_dpdk_enable=Tenant-networks-only: Enables OVS-DPDK in Mode 1.
+# ovs_dpdk_enable=Tenant-and-External-networks: Enables OVS-DPDK in Mode 2.
+ovs_dpdk_enable=None
```
cdearborn: Can we use False to indicate disabled for consistency with the other *_enable parameters?

src/deploy/osp_deployer/settings/sample_xsp_profile.ini
```diff
@@ -171,6 +171,30 @@ StorageBond1Interface1=em2
...
+
+# Comma separated list of roles on which OVS-DPDK needs to be enabled
+# Only supported value is "Compute".
+ovs_dpdk_role_list=Compute
+
+# Cores distribution policy to be used. Only supported value is "Balanced".
+ovs_dpdk_policy=Balanced
+# Refer to guide for more details about each parameter.
+
+hpg_enable=False
```
cdearborn: The hpg_enable through numa_flavor_name_list parameter should be removed from this patch since they are being added under the hpg/numa patch.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,6 +171,31 @@ StorageBond1Interface1=em2
...
+
+# Comma separated list of roles on which OVS-DPDK needs to be enabled
+# Only supported value is "Compute".
+ovs_dpdk_role_list=Compute
+
+# Cores distribution policy to be used. Only supported value is "Balanced".
+ovs_dpdk_policy=Balanced
+# Refer to guide for more details about each parameter.
+
+hpg_enable=False
```
cdearborn: The hpg_enable through numa_flavor_name_list parameter should be removed from this patch since they are being added under the hpg/numa patch.

src/deploy/osp_deployer/director.py
```diff
@@ -706,17 +709,44 @@ def configure_dhcp_server(self):
     def setup_networking(self):
         logger.debug("Configuring network settings for overcloud")
 
+        if self.settings.enable_ovs_dpdk:
```
cdearborn: It looks like these 3 lines are unnecessary because both mode and compute_file_name are set below on lines 731/732.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,238 @@
...
+  ComputeBond1Interface1:
+    default: 'em2'
+    description: Bond 1 interface 1 name
+    type: string
+  ComputeBond1Interface2:
+    default: 'p3p2'
+    description: Bond 1 interface 2 name
+    type: string
+  ComputeBondInterfaceOptions:
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
```
cdearborn: The above line should be:
default: 'mode=802.3ad ...'

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,238 @@
...
+  ComputeBond1Interface1:
+    default: 'em2'
+    description: Bond 1 interface 1 name
+    type: string
+  ComputeBond1Interface2:
+    default: 'p3p2'
+    description: Bond 1 interface 2 name
+    type: string
+  ComputeBondInterfaceOptions:
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
```
cdearborn: The above line should be:
default: 'mode=802.3ad ...'

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,244 @@
...
+  ComputeBond1Interface1:
+    default: 'em2'
+    description: Bond 1 interface 1 name
+    type: string
+  ComputeBond1Interface2:
+    default: 'p3p2'
+    description: Bond 1 interface 2 name
+    type: string
+  ComputeBondInterfaceOptions:
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
```
cdearborn: The above line should be:
default: 'mode=802.3ad ...'

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,244 @@
...
+  ComputeBond1Interface1:
+    default: 'em2'
+    description: Bond 1 interface 1 name
+    type: string
+  ComputeBond1Interface2:
+    default: 'p3p2'
+    description: Bond 1 interface 2 name
+    type: string
+  ComputeBondInterfaceOptions:
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
```
cdearborn: The above line should be:
default: 'mode=802.3ad ...'

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,238 @@
...
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
```
cdearborn: The BondInterfaceOptions parameter should be removed (lines 113-117) since it is no longer used.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,238 @@
...
+    type: comma_delimited_list
+  EC2MetadataIp: # Override this via parameter_defaults
+    description: The IP address of the EC2 metadata server.
+    type: string
+  ComputeBond0Interface1:
+    default: 'em1'
+    description: Bond 0 interface 1 name
+    type: string
+  ComputeBond0Interface2:
+    default: 'p3p1'
```
cdearborn: Is there a compelling reason for changing the NIC PCI slot from 1 to 3?

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,238 @@
...
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
+    default: ''
+    description: The bonding_options string for the bond interface. Set things
+                 like mode=802.3ad using this option.
+    type: string
+  OvsDpdkBondInterfaceOptions:
+    default: ''
```
cdearborn: It seems like we should set the default above correctly: 'bond_mode=balance-tcp lacp=active'

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,238 @@
...
+    type: comma_delimited_list
+  EC2MetadataIp: # Override this via parameter_defaults
+    description: The IP address of the EC2 metadata server.
+    type: string
+  ComputeBond0Interface1:
+    default: 'em1'
+    description: Bond 0 interface 1 name
+    type: string
+  ComputeBond0Interface2:
+    default: 'p3p1'
```
cdearborn: Is there a compelling reason for changing the NIC PCI slot from 1 to 3?

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,238 @@
...
+    type: string
+  ComputeBond1Interface2:
+    default: 'p3p2'
+    description: Bond 1 interface 2 name
+    type: string
+  ComputeBondInterfaceOptions:
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
```
cdearborn: The BondInterfaceOptions parameter should be removed.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,238 @@
...
+                      - - {get_param: ControlPlaneIp}
+                        - {get_param: ControlPlaneSubnetCidr}
+              routes:
+                -
+                  ip_netmask: 169.254.169.254/32
+                  next_hop: {get_param: EC2MetadataIp}
+                -
+                  default: true
+                  next_hop: {get_param: ControlPlaneDefaultRoute}
+
```
cdearborn: The above line has several spaces on it.  Can you remove the spaces and leave the blank line?

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,244 @@
...
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
```
cdearborn: The BondInterfaceOptions parameter should be removed.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,238 @@
...
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
+    default: ''
+    description: The bonding_options string for the bond interface. Set things
+                 like mode=802.3ad using this option.
+    type: string
+  OvsDpdkBondInterfaceOptions:
+    default: ''
```
cdearborn: It seems like we should set the default above correctly: 'bond_mode=balance-tcp lacp=active'

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,244 @@
...
+    default: ''
+    description: The bonding_options string for the bond interface. Set things
+                 like mode=802.3ad using this option.
+    type: string
+  OvsDpdkBondInterfaceOptions:
+    default: ''
```
cdearborn: It seems like we should set the default above correctly: 'bond_mode=balance-tcp lacp=active'

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,244 @@
...
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
+    default: ''
+    description: The bonding_options string for the bond interface. Set things
+                 like mode=802.3ad using this option.
+    type: string
+  OvsDpdkBondInterfaceOptions:
+    default: ''
```
cdearborn: It seems like we should set the default above correctly: 'bond_mode=balance-tcp lacp=active'

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,244 @@
...
+    type: string
+  ComputeBond1Interface2:
+    default: 'p3p2'
+    description: Bond 1 interface 2 name
+    type: string
+  ComputeBondInterfaceOptions:
+    default: '802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1'
+    description: The bonding_options string for the bond interface.
+    type: string
+  BondInterfaceOptions:
```
cdearborn: The BondInterfaceOptions parameter should be removed.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,6 +171,31 @@ StorageBond1Interface1=em2
 StorageBond1Interface2=p2p2
 StorageBondInterfaceOptions=mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
 
+[Dell NFV Settings]
```
cdearborn: There is a second Dell NFV Settings section below at https://github.com/dsp-jetpack/JetPack/pull/28/files#diff-03907ea6ba7811b5a8d4acb7b8c60701L255.
One of these sections needs to be removed.

cdearborn:
The copyright dates on all the new files should be updated to 2018

src/pilot/templates/neutron-ovs-dpdk.yaml
```diff
@@ -0,0 +1,55 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## A Heat environment that can be used to deploy DPDK with OVS
+resource_registry:
+  OS::TripleO::Compute::Net::SoftwareConfig: ./nic-configs/compute-ovs-dpdk-mode1.yaml
```
cdearborn: The path in the above line is wrong, but the whole line should be removed since the NIC configuration is now stored in the nic_environment.yaml.

src/pilot/templates/neutron-ovs-dpdk.yaml
```diff
@@ -0,0 +1,55 @@
...
+  NeutronDpdkMemoryChannels: "4"
+  # NeutronDpdkSocketMemory
+  NeutronDpdkSocketMemory: "2048,2048"
+  # NeutronDpdkDriverType
+  NeutronDpdkDriverType: "vfio-pci"
+  # Datapath type for ovs bridges
+  NeutronDatapathType: "netdev"
+  # The vhost-user socket directory for OVS
+  NeutronVhostuserSocketDir: "/var/run/openvswitch"
+  DpdkInterfaces: 'p2p1,p3p1'
```
cdearborn: The above doesn't look right since p3p1 is a leg of bond0, and p2p1 isn't in the compute NIC config files at all...

src/pilot/templates/post-install.yaml
```diff
@@ -0,0 +1,60 @@
...
+  ExtraConfig:
+    type: OS::Heat::SoftwareConfig
+    properties:
+      group: script
+      config: |
+        #!/bin/bash
+
+        set -x
+        if [[ $HOSTNAME =~ "compute" ]] ; then
+          VCPUPINSET=$(crudini --get /root/parameters.ini DEFAULT novacpus)
```
cdearborn: Both the NUMA/Hugepages patch and this one should set the vcpu_pin_set in the same way.  I like the tripleo method of doing it since there are fewer moving parts.

src/pilot/ovs_dpdk_setup.py
```diff
@@ -0,0 +1,146 @@
...
+    else:
+        raise Exception('ERROR: Mode ' + mode + ' is not supported.'
+                        'Only supported values are 1 and 2.')
+    setup_ovs_dpdk_environment(nics, args.config_file, args.env_file, mode)
+
+if __name__ == "__main__":
+    try:
+        main()
+    except Exception as e:
+        print str(e)
```
cdearborn: The above print statement should be replaced with a logging statement.

cdearborn:
I diffed the mode1 and mode2 4_port NIC configs, and they are identical.  Since they are the same, we should eliminate one of them and call it something like "compute-oovs-dpdk.yaml".
The same is true for the mode 1 and mode 2 5_port NIC configs.

src/deploy/osp_deployer/director.py
jproque-dell: Agree. Removed

src/deploy/osp_deployer/settings/sample_csp_profile.ini
jproque-dell: Removed second one. Kept only one as suggested.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
jproque-dell: Agree. Removed.

src/pilot/ovs_dpdk_setup.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
```
cdearborn: I really feel that this piece of code should be eliminated.  This is because this code makes it so that DPDK can only be used with R630 and R730 servers, and it also hard codes the NICs for these servers into one configuration.  We have had requests to make the solution more flexible, and this piece of code moves us in the opposite direction.

src/deploy/osp_deployer/director.py
```diff
@@ -706,17 +709,44 @@ def configure_dhcp_server(self):
...
         # leave the undercloud intact but want to redeploy
         # with a different config
         self.upload_file(self.settings.static_ips_yaml, static_ips_yaml)
         self.upload_file(self.settings.static_vip_yaml, static_vip_yaml)
+        self.upload_file(self.settings.neutron_ovs_dpdk_yaml,
+                         neutron_ovs_dpdk_yaml)
 
         self.setup_nic_configuration()
 
+        if self.settings.enable_ovs_dpdk is True:
```
cdearborn: If we agree that we should not restrict the DPDK feature to a specific NIC configuration on only 2 server models, then this code (729-749) should also be removed.

cdearborn:
I think that instead of modifying the existing NIC configs that we should create 2 new dpdk 4_port and 5_port nic configs.  This would allow us to replace the changeme tokens in the added compute NIC config files with actual variable names and then add those variables to a new nic_environment.yaml.  Taking this approach will take advantage of the standard NIC configuration mechanism.

src/pilot/templates/first-boot-ovs-dpdk.yaml
```diff
@@ -0,0 +1,450 @@
...
+    properties:
+      parts:
+      - config: {get_resource: wipe_disk}
+      - config: {get_resource: generate_parameters}
+      - config: {get_resource: set_ovs_config}
+      - config: {get_resource: set_dpdk_params}
+      - config: {get_resource: install_tuned}
+      - config: {get_resource: compute_kernel_args}
+
+  wipe_disk:
```
cdearborn: Is there some way to pull in the existing wipe_disk code instead of duplicating it here?

src/pilot/ovs_dpdk_setup.py
```diff
@@ -0,0 +1,146 @@
...
+        interfaces = "'" + ",".join(nics) + "'"
+
+    cmds = [
+        'sed -i "s|changeme1|' + nics[0] + '|" ' + config_file,
+        'sed -i "s|changeme2|' + nics[1] + '|" ' + config_file,
+        'sed -i "s|changeme3|' + nics[2] + '|" ' + config_file,
+        'sed -i "s|changeme4|' + nics[3] + '|" ' + config_file,
+        'sed -i "s|Compute::Net::SoftwareConfig:.*|' +
+        'Compute::Net::SoftwareConfig: ' + config_file + '|" ' + env_file,
+        'sed -i "s|DpdkInterfaces:.*|DpdkInterfaces: ' +
```
cdearborn: The line above and following, we need to keep.  Instead of using this mechanism, it seems like we should do something similar to how we modify dell-environment.yaml in automation.

src/pilot/ovs_dpdk_setup.py
jproque-dell: Agree. Done

src/deploy/osp_deployer/settings/sample_xsp_profile.ini
jproque-dell: Yes. Changed.

src/pilot/ovs_dpdk_setup.py
jproque-dell: Removing this hardcoding is part of the planned activities. It's currently in-progress.

src/pilot/templates/first-boot-ovs-dpdk.yaml
jproque-dell: I'm not aware of any technical method to call this wipe_disk code outside of its underlying template. I don't think we can use it as a regular python method if that's what you meant.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Done.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
jproque-dell: Nope. Probably a typo. Fixed.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Nope. Probably a typo. Fixed.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Agree. Done.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
jproque-dell: Agree. Done.

src/pilot/templates/neutron-ovs-dpdk.yaml
jproque-dell: p3p1 is not supposed to be a leg of Bond0, it's fixed now. p2p1 and p3p1 DpdkInterfaces are part of the hardcoding piece of code.

src/pilot/templates/neutron-ovs-dpdk.yaml
jproque-dell: Agree. Removed.

src/pilot/ovs_dpdk_setup.py
jproque-dell: Would it be better to check if the server that we have chosen to work with is supported by a kind of support matrix? Maybe we can enforce this check by use of our profiles.

src/deploy/osp_deployer/settings/sample_xsp_profile.ini
jproque-dell: Agree. Removed.

src/deploy/osp_deployer/settings/config.py
```diff
@@ -240,6 +240,34 @@ def __init__(self, settings_file):
...
+            raise AssertionError('Only supported values for '
+                                 'ovs_dpdk_enable are false, '
+                                 'Tenant-networks-only and '
+                                 'Tenant-and-External-networks')
+        if self.enable_ovs_dpdk:
+            logger.info("OVS_DPDK is enabled with mode " +
+                        str(self.ovs_dpdk_mode) + ".")
+        else:
+            logger.info("OVS_DPDK is disabled.")
+
```
gaelrehault: enable_ovs_dpdk  & ovs_dpdk_enable are essentially the same thing, could we just use ovs_dpdk_enable ?
if self.enable_ovs_dpdk can change to self.ovs_dpdk_enable.lower() != 'false':

gaelrehault:
There should a check on features dependencies associated with ovs-dpdk . Profile already enforces that, but in case of a custom profile  we still want to make sure that if ovs-dpdk is enabled, numa and hp also are. ( in settings_sanity.py possibly , but also deploy-overcloud.py for manual deployments)

src/pilot/ovs_dpdk_setup.py
cdearborn: Well, I think we want users to be able to use this with any hardware.  It may not work if it's hardware we haven't tested with, but chances are good that it will work.

src/pilot/templates/neutron-ovs-dpdk.yaml
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2017 Dell Inc. or its subsidiaries.
```
cdearborn: Can you update the copyright date?

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,249 @@
+# Copyright (c) 2016-2018 Dell Inc. or its subsidiaries.
```
cdearborn: I think we should create a new nic-config for this instead of trying to add to an existing NIC config.

src/pilot/templates/nic-configs/4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,243 @@
+# Copyright (c) 2016-2018 Dell Inc. or its subsidiaries.
```
cdearborn: I think we should create a new nic-config for this instead of trying to add to an existing NIC config.

src/pilot/templates/nic-configs/4_port/nic_environment.yaml
```diff
@@ -45,3 +45,13 @@ parameter_defaults:
   StorageBond1Interface2: p4p2
   # The bonding mode to use for storage nodes
   StorageBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+
+  # CHANGEME: Change the interface names in the following lines for the
```
cdearborn: I think we should create a new nic-config for this instead of trying to add to an existing NIC config.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -0,0 +1,255 @@
+# Copyright (c) 2016-2018 Dell Inc. or its subsidiaries.
```
cdearborn: I think we should create a new nic-config for this instead of trying to add to an existing NIC config.

src/pilot/templates/nic-configs/5_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -0,0 +1,250 @@
+# Copyright (c) 2016-2018 Dell Inc. or its subsidiaries.
```
cdearborn: I think we should create a new nic-config for this instead of trying to add to an existing NIC config.

src/pilot/templates/nic-configs/5_port/nic_environment.yaml
```diff
@@ -51,3 +51,13 @@ parameter_defaults:
   StorageBond1Interface2: p2p2
   # The bonding mode to use for storage nodes
   StorageBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+
+  # CHANGEME: Change the interface names in the following lines for the
```
cdearborn: I think we should create a new nic-config for this instead of trying to add to an existing NIC config.

src/deploy/osp_deployer/settings/config.py
jproque-dell: ovs_dpdk_enable can have three states, it's not a boolean. It can accept Tenant-networks-only which is mode 1, Tenant-and-External-networks which is mode 2 or false which disable the feature. That way it exposes the feature in a more user-friendly way. On the other hand, enable_ovs_dpdk determines only if the feature is enabled or not,regardless of the mode. Besides, this one is a boolean.

src/pilot/templates/neutron-ovs-dpdk.yaml
jproque-dell: Done.

src/pilot/templates/post-install.yaml
jproque-dell: I agree, and this is part of the future way of implementation. Both NUMA and Hugepages will be handled by former calsoft code, and we need now to think how we can eliminate this dependency interlock and put all of that together.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,6 +173,19 @@ StorageBond1Interface1=em2
...
+# R630 should use 'p2p1', 'p3p1', 'p2p2', 'p3p2'
+# R730 should use 'p4p1', 'p5p1', 'p4p2', 'p5p2'
+# The following lines must be uncommented only if OVS-DPDK is set to enable 
+# See Dell NFV Settings section -> option ovs_dpdk_enable
+ComputeOvsDpdkInterface1=p2p1
+ComputeOvsDpdkInterface2=p3p1
+ComputeOvsDpdkInterface3=p2p2
+ComputeOvsDpdkInterface4=p3p2
+BondInterfaceOvsOptions=bond_mode=balance-tcp lacp=active
+
```
gaelrehault: i m a little confused by the above nics, our default nics enumeration for computes is em1/p1p1 em2/p1p2. This is the default defined in both our sample and nic config files, where do those p2p* & p4p* & p5p* come from ?

jproque-dell: Those bond are used only for Dpdk, remember the hardcoding in the code? regular bonding on compute remain the same aka em1/p1p1 em2/p1p2.

src/pilot/templates/nic-configs/5_port/nic_environment.yaml
jproque-dell: Agree. new nic-config dpdk_4_port and dpdk_5_port have been created.

src/pilot/templates/nic-configs/4_port/nic_environment.yaml
jproque-dell: Agree. new nic-config dpdk_4_port and dpdk_5_port have been created.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,19 +173,48 @@ StorageBond1Interface1=em2
...
 #Enter value of enable_hpg as True/False for HugePages
-hpg_enable=True
+hpg_enable=true
 #User should give this parameter in same format.
 #Supported values for hpg_size(Size of hugepages) is 2MB and 1 GB.
 #For 2MB (# of hugepages), hpg_num is 49152
 #For 1GB (# of hugepages), hpg_num is 96
 hpg_size=2MB
 
 #Enter value of enable_numa as True/False for NUMA
```
cdearborn: Should change "True/False" to "true/false" in the above line for consistency with the xsp profile.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,19 +173,48 @@ StorageBond1Interface1=em2
...
+ComputeOvsDpdkInterface2=p3p1
+ComputeOvsDpdkInterface3=p2p2
+ComputeOvsDpdkInterface4=p3p2
+BondInterfaceOvsOptions=bond_mode=balance-tcp lacp=active
+
+
 [Dell NFV Settings]
 #Provide NFV features here.
 
 #Enter value of enable_hpg as True/False for HugePages
```
cdearborn: Should change "True/False" to "true/false" in the above line for consistency with the xsp profile.

src/deploy/osp_deployer/profiles/xsp.json
```diff
@@ -50,17 +63,17 @@
...
             "deploy_overcloud_debug": {
                "valid_values": [
                   "true",
                   "false"
                ]
             },
-         "use_ipmi_driver":"",
-         "use_in_band_introspection":""
+            "use_ipmi_driver":"",
+            "use_in_band_introspection":""
```
cdearborn: Probably want to change :"" to : "" on the 2 lines above for consistency in formatting.

src/deploy/osp_deployer/profiles/xsp.json
```diff
@@ -4,13 +4,26 @@
    "associated_settings": [
       {
          "Dell NFV Settings": {
-            "hpg_enable": "",
-            "numa_enable": ""
+            "ovs_dpdk_enable": {
```
cdearborn: Can you add the sections for ovs_dpdk_role_list and ovs_dpdk_policy to this section?

src/pilot/templates/nic-configs/dpdk_4_port/nic_environment.yaml
```diff
@@ -0,0 +1,57 @@
...
+  # CHANGEME: Change the interface names in the following lines to include in
+  #           the storage nodes bonds
+  StorageBond0Interface1: em1
+  StorageBond0Interface2: p4p1
+  StorageBond1Interface1: em2
+  StorageBond1Interface2: p4p2
+  # The bonding mode to use for storage nodes
+  StorageBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+
+  # CHANGEME: Change the interface names in the following lines for the
```
cdearborn: Can you move these ComputeOvsDpdkInterface* lines up to the compute section?

src/pilot/templates/nic-configs/dpdk_4_port/nic_environment.yaml
```diff
@@ -0,0 +1,57 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource_registry:
+  OS::TripleO::Controller::Net::SoftwareConfig: ./controller.yaml
+  OS::TripleO::DellCompute::Net::SoftwareConfig: ./compute.yaml
```
cdearborn: Can you change the above line so that it references either compute-ovs-dpdk-mode1.yaml  compute-ovs-dpdk-mode2.yaml since compute.yaml doesn't exist for this NIC config?

src/pilot/templates/nic-configs/dpdk_5_port/nic_environment.yaml
```diff
@@ -0,0 +1,63 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource_registry:
+  OS::TripleO::Controller::Net::SoftwareConfig: ./controller.yaml
+  OS::TripleO::DellCompute::Net::SoftwareConfig: ./compute.yaml
```
cdearborn: Can you change the above line so that it references either compute-ovs-dpdk-mode1.yaml compute-ovs-dpdk-mode2.yaml since compute.yaml doesn't exist for this NIC config?

src/pilot/templates/nic-configs/dpdk_5_port/nic_environment.yaml
```diff
@@ -0,0 +1,63 @@
...
+  StorageBond0Interface2: p2p1
+  StorageBond1Interface1: em2
+  StorageBond1Interface2: p2p2
+  # The bonding mode to use for storage nodes
+  StorageBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+
+  # CHANGEME: Change the interface names in the following lines for the
+  # compute nodes provisioning interface and to include in the compute
+  # DPDK bonds
+  ComputeOvsDpdkInterface1: p2p1
```
cdearborn: Can you move these ComputeOvsDpdkInterface* lines up to the compute section?

src/pilot/ovs_dpdk_setup.py
```diff
@@ -0,0 +1,142 @@
...
+        config_file), config_file + " file does not exist"
+    assert os.path.isfile(
+        env_file), env_file + " file does not exist"
+    if mode == 2:
+        interfaces = "'" + ",".join(nics[0:2]) + "'"
+    else:
+        interfaces = "'" + ",".join(nics) + "'"
+
+    cmds = [
+        'sed -i "s|Compute::Net::SoftwareConfig:.*|' +
```
cdearborn: I'm pretty sure that the ab

src/pilot/ovs_dpdk_setup.py
```diff
@@ -0,0 +1,142 @@
...
+    assert os.path.isfile(
+        config_file), config_file + " file does not exist"
+    assert os.path.isfile(
+        env_file), env_file + " file does not exist"
+    if mode == 2:
+        interfaces = "'" + ",".join(nics[0:2]) + "'"
+    else:
+        interfaces = "'" + ",".join(nics) + "'"
+
+    cmds = [
```
cdearborn: I'm pretty sure that the next 2 lines don't do anything and should be removed since "Compute::Net::SoftwareConfig" does not appear in neutron-ovs-dpdk.yaml.

src/deploy/osp_deployer/profiles/xsp.json
jproque-dell: Sure, my fault! It's added now. I also changed the order of the options in the csp.json file to make it consistent.

src/pilot/templates/nic-configs/dpdk_4_port/nic_environment.yaml
jproque-dell: Moved.

src/pilot/templates/nic-configs/dpdk_4_port/nic_environment.yaml
jproque-dell: Changed.

src/pilot/templates/nic-configs/dpdk_5_port/nic_environment.yaml
jproque-dell: Moved.

src/deploy/osp_deployer/settings/config.py
```diff
@@ -1,528 +1,561 @@
...
+            self.sah_bond_opts = nics_settings['sah_bond_opts']
+
+        # This particular section has been moved right after the nics_settings
+        # section in order to catch the mode used by the nic-environment file
+        dellnfv_settings = self.get_settings_section(
+            "Dell NFV Settings")
+        self.ovs_dpdk_enable = dellnfv_settings[
+            'ovs_dpdk_enable']
+        self.ovs_dpdk_role_list = dellnfv_settings[
+            'ovs_dpdk_role_list']
```
gaelrehault: I see this here & in both the  ini and the json files, but dont see it being used anywhere ? If so could we remove all references to it ?

cdearborn:
We should delete the dpdk_4_port and dpdk_5_port nic_configs since these are no longer needed.

src/deploy/osp_deployer/director.py
```diff
@@ -846,6 +855,57 @@ def setup_nic_configuration(self):
...
+            interfaces = "'" + ",".join(dpdk_interfaces[0:2]) + "'"
+        else:
+            interfaces = "'" + ",".join(dpdk_interfaces) + "'"
+
+        # Build up the sed command to perform variable substitution
+        # in the neutron-ovs-dpdk.yaml (dpdk environment)
+        cmds.append('sed -i "s|DpdkInterfaces:.*|DpdkInterfaces: ' +
+                    interfaces + '|" ' + env_file)
+
+        nic_config_dirname = os.path.dirname(self.settings.nic_env_file)
```
cdearborn: I don't think lines 886-903 are needed since the nic_env_file will already be pointing at the correct compute.yaml file.

src/deploy/osp_deployer/settings/config.py
cdearborn: the same is true for both ovs_dpdk_role_list and ovs_dpdk_policy

src/deploy/osp_deployer/settings/config.py
```diff
@@ -1,528 +1,561 @@
...
+        self.ovs_dpdk_policy = dellnfv_settings[
+            'ovs_dpdk_policy']
+        self.enable_ovs_dpdk = False
+        if self.ovs_dpdk_enable.lower() == 'false':
+            pass
+        elif self.ovs_dpdk_enable.lower() == 'true':
+            self.enable_ovs_dpdk = True
+            for each in re.split(r'[_/]', self.nic_env_file):
```
cdearborn: Hrm...  I don't think the above line is working.  It will try to split the value of nic_env_file up by occurrences of "_/", but nic_env_file looks like: "dpdk_mode1_4_port/nic_environment.yaml", which doesn't have any occurrences of that, so each will be "dpdk_mode1_4_port/nic_environment.yaml".  The line below will succeed in finding "mode", but each[-1:] will return the "l" from "yaml".
I think you want something like:
match=re.search('mode(\d)', self.nic_env_file)
if match is None:
    raise some_exception_here_due_to_no_match
self.ovs_dpdk_mode = match.group(1)

src/deploy/osp_deployer/settings/config.py
```diff
@@ -1,528 +1,561 @@
...
+                if each.find('mode') != -1:
+                    self.ovs_dpdk_mode = each[-1:]
+        else:
+            raise AssertionError('Only supported values for '
+                                 'ovs_dpdk_enable are true or false. ')
+        if self.enable_ovs_dpdk:
+            mode = str(self.ovs_dpdk_mode)
```
cdearborn: If you make the above change, then mode will be a string already, so the above line won't be needed.

src/deploy/osp_deployer/director.py
jproque-dell: I don't think so. The name of the compute.yaml will changed based on the mode enabled. For example if mode 1 is chosen then it needs to point to compute-ovs-dpdk-mode1.yaml, if mode 2 is chosen then compute-ovs-dpdk-mode2.yaml will be picked up.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -171,19 +176,45 @@ StorageBond1Interface1=em2
 StorageBond1Interface2=p2p2
 StorageBondInterfaceOptions=mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
 
+# Note: the following interfaces need to be changed 
+# as per the server model
+# R630 should use 'p2p1', 'p3p1', 'p2p2', 'p3p2'
+# R730 should use 'p4p1', 'p5p1', 'p4p2', 'p5p2'
+# The following lines must be uncommented only if OVS-DPDK is set to enable
```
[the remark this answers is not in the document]
jproque-dell: Indeed, fixed.

src/deploy/osp_deployer/settings/config.py
jproque-dell: the line does work actually.
see below:
>>> re.split(r'[/_]', 'dpdk_mode1_4_port/nic_environment.yaml')
['dpdk', 'mode1', '4', 'port', 'nic', 'environment.yaml']

src/deploy/osp_deployer/settings/config.py
jproque-dell: The ovs_dpdk_mode was already a string, this line is therefore useless.

src/deploy/osp_deployer/settings/config.py
jproque-dell: Agree. It was there for future implementation. Removed.

src/deploy/osp_deployer/settings/config.py
cdearborn: Ah, I see.  It's splitting on / OR _.

jproque-dell: Bingo!

PR #30, opened 2018-03-16
Fixed mistake in bonding options

cdearborn:
The bonding mode requires "mode=802.3ad" instead of just "802.3ad".


cdearborn:
This patch is clean.  I accidentally included "error" in the PR title.

PR #32, opened 2018-03-20
Updates with configurations for 4 and 5 nic samples

rperryman:
Added the 5-NIC configuration that would support 25G.
Moved the 10G Configuration to a 4-NIC-10G no changes just file names. 
Left Old configuration in place untill everything is all settled down.


cdearborn:
Looks good to me.

heydog15:
Only thing I see is that the 4-nic workbook has stamp names on the logical layout page under the Vlan section.  Has R7-3048 on 2 rows.  Also the DNS for Floating net looks out of range unless routers are setup.

jtwill98:
Lets address Kurt's eagle eye catches before merging.

rperryman:
Pushed updated 4-Nic spreadsheet.

heydog15:
Looks good.

PR #33, opened 2018-03-22
MTU Size support for JS-10.2 in both profiles 

tahir696:
Added new flags/parameters to the .ini file.
Modification in config.py to read new flags.
Added checks for MTU val <592 and >9216  in settings_sanity.py.
Added check mtu value function in checkpoints.py.
Changed sah.py to modify ks and .cfg files.
Addition of new parameters in environment.yaml and nic-config files for 4 port and 5 port config.


src/deploy/osp_deployer/settings_sanity.py
```diff
@@ -128,6 +135,23 @@ def check_ipmi_to_nodes(self):
...
+    def check_mtu_of_networks(self):
+        shouldbbevalidmtu = [
+            'mtu_size_global_default',
+            'public_api_network_mtu',
+            'private_api_network_mtu',
+            'storage_network_mtu',
+            'provisioning_network_mtu',
+            'storage_cluster_network_mtu',
+            'tenant_tunnel_network_mtu',
+            'tenant_network_mtu',
```
cdearborn: Shouldn't we have checks here for floating_ip_network_mtu and management_network_mtu?

src/mgmt/deploy-dashboard-vm.py
```diff
@@ -339,26 +341,33 @@ def create_kickstart(ks_filename, cfg_filename):
...
                          format(
                              tokens[0],
                              tokens[1],
                              tokens[2],
                              hostname,
                              gateway,
                              nameservers,
+                             tokens[3],
                              ks_include))
+                ks.write("echo eth0='{}' >> {}\n".
```
cdearborn: Can we change the variable name to something like eth0_mtu or something like that?

src/mgmt/deploy-director-vm.sh
```diff
@@ -166,6 +168,10 @@ EOFPW
   done
 
   echo "GATEWAY=${Gateway}" >> /etc/sysconfig/network
+  echo "MTU=${eth0}" >> /etc/sysconfig/network-scripts/ifcfg-eth0
```
cdearborn: Can we change the mtu variable names for the interfaces to something like eth0_mtu?

src/mgmt/deploy-director-vm.sh
```diff
@@ -70,7 +70,7 @@ EOFKS
 {
```
cdearborn: Can you add execute permissions back to this file?

src/pilot/templates/nic-configs/5_port/controller.yaml
```diff
@@ -168,11 +201,13 @@ resources:
...
+              mtu: {get_param: DefaultBondMTU}
               members:
                 -
                   type: linux_bond
                   name: bond0
                   bonding_options: {get_param: ControllerBondInterfaceOptions}
+                  mtu: {get_param: DefaultBondMTU}
                   members:
                     -
                       type: interface
```
cdearborn: I think we missed setting mtu on the two controller interfaces below.

src/pilot/templates/nic-configs/5_port/nic_environment.yaml
```diff
@@ -28,7 +28,7 @@ parameter_defaults:
   ControllerBond1Interface1: em2
   ControllerBond1Interface2: p3p2
   # The bonding mode to use for controller nodes
```
cdearborn: The MTU variables that were added to the NIC configuration files should also be added here.

src/pilot/templates/nic-configs/4_port/nic_environment.yaml
```diff
@@ -26,7 +26,7 @@ parameter_defaults:
   ControllerBond1Interface1: em2
   ControllerBond1Interface2: p1p2
   # The bonding mode to use for controller nodes
```
cdearborn: The MTU variables that were added to the NIC configuration files should also be added here.

src/pilot/templates/network-environment.yaml
```diff
@@ -105,7 +107,34 @@ parameter_defaults:
   # CHANGEME: Change the following to the VLAN ID to use on the
   # External network
   ExternalNetworkVlanID: 190
-
```
cdearborn: The MTU variables below (lines 110-133) should be in the nic_environment.yaml files, and should not appear here.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -81,6 +81,21 @@ tenant_tunnel_network_vlanid=130
 tenant_vlan_range=201:250
 
 
+[MTU Settings]
```
cdearborn: I think all of the variables below should be moved into the [Nodes Nics and Bonding Settings] section, and the *_network_mtu variables should be replaced with the actual variable names in the NIC config files.

src/deploy/osp_deployer/settings/sample_xsp_profile.ini
```diff
@@ -80,6 +80,21 @@ tenant_tunnel_network_vlanid=130
 # Nova Private network details
 tenant_vlan_range=201:250
 
+[MTU Settings]
```
cdearborn: I think all of the variables below should be moved into the [Nodes Nics and Bonding Settings] section, and the *_network_mtu variables should be replaced with the actual variable names in the NIC config files.

src/deploy/osp_deployer/settings/config.py
```diff
@@ -159,6 +159,40 @@ def __init__(self, settings_file):
         self.tenant_tunnel_vlanid = network_settings[
             'tenant_tunnel_network_vlanid']
         self.tenant_vlan_range = network_settings['tenant_vlan_range']
+        mtu_settings = self.get_settings_section(
```
cdearborn: When the MTU variables are moved into the [Nodes Nics and Bonding Settings], most of the code will go away, and the remainder will move elsewhere.

src/deploy/osp_deployer/settings_sanity.py
tahir696: sure will add the checks

src/pilot/templates/nic-configs/5_port/nic_environment.yaml
tahir696: The "nic-environment.yaml" file contains settings that are related to the NIC names and bonding options, where as MTU is related to Networks just like network IP ranges and VLAN IDs, so it should be in "network-environment.yaml" files. Moreover, "nic-environment.yaml" is part of the customizable NIC configuration scheme. Hence, it should only contain variables/parameters that are affecting the overcloud NIC configs only. Its not a good idea to make MTU a part of that as we make changes in the SAH, CEPH dashboard VM and director VM based on these settings. 

Also we have "NeutronGlobalPhysnetMtu: 1500" which is used to set the Global value for neutron to create the overcloud tenant networks. This parameter name should stay as it is and its value comes from the "tenant_network_mtu" in the ini file. So, we cannot have it a part of a file that is entirely customizable. 

MTU setting for floating network is done through the extra config i.e. neutron::plugins::ml2::physical_network_mtus: physext:1500, which we cannot make a part of "nic-environment.yaml". 

So as per our understanding network-environment.yaml is the appropriate place to keep these settings.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
tahir696: this was the decision made by sumedh and Randy to keep a separate section/stanza for mtu settings

src/deploy/osp_deployer/settings/sample_xsp_profile.ini
tahir696: this was the decision made by sumedh and Randy to keep a separate section/stanza for mtu settings

src/mgmt/deploy-dashboard-vm.py
tahir696: will change it to eth0_mtu

src/mgmt/deploy-director-vm.sh
tahir696: will change it to eth0_mtu

src/pilot/templates/nic-configs/5_port/controller.yaml
tahir696: will add the mtu settings for interfaces as well

gaelrehault:
Can you please update to pick the Numa/HugePages PR that was merged yesterday

src/pilot/templates/nic-configs/4_port/controller.yaml
```diff
@@ -167,56 +200,67 @@ resources:
...
                       primary: true
                     -
                       type: interface
                       name: {get_param: ControllerBond0Interface2}
+                      mtu: {get_param: DefaultBondMTU}
                 -
                   type: vlan
                   device: bond0
-                  vlan_id: {get_param: TenantNetworkVlanID}
+                  vlan_id: {get_param: TenantNetworkVlanID
```
cdearborn: It looks like a closing "}"was accidentally deleted at the end of the above line.

src/deploy/osp_deployer/sah.py
```diff
@@ -183,6 +185,67 @@ def update_kickstart_usb(self):
                                       'prov_netmask="' +
                                       sets.provisioning_netmask +
                                       '"')
+# mtu_settings
```
cdearborn: Can you indent the above line so it matches with the line below?

tahir696:
I am done with the changes can you please accept/merge the pull request

gaelrehault:
Both profiles are set to use a 1500 MTU accross the networks here, which is the default value used anyway, so given the current implementation and profiles, MTU values can only be changed if using a custom profile , is this intentional ?

alp317:
Both profiles default to 1500, but as required it can be changed in both profiles. I am not sure on what classifies as a custom profile.

gaelrehault:
sorry you re correct, those values aren t set in the json

PR #35, opened 2018-03-22
Detecting boss card and select os volume

digambar15:
1. List all the controllers and find boss controller from it
2. Find out the boss card from physical disks using controller model id
4. Calculate the boss card size in GB
5. Pass to select_os_volume method.
6. When we pass the voume size to that method, method directly submit to ironic client

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -1094,9 +1094,9 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
 
             raid_physical_disk_ids = []
 
-            # Look for a RAID of any type other than RAID0 and assume we want to
-            # install the OS on that volume.  The first non-RAID0 found will be
-            # used.
+            # Look for a RAID of any type other than RAID0 and assume we want
+            # to install the OS on that volume.  The first non-RAID0 found
+            # will be used.
             raid_size_gb = 0
             for virtual_disk_doc in virtual_disk_docs:
                 fqdd = get_fqdd(virtual_disk_doc, DCIM_VirtualDiskView)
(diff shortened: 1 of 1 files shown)


src/pilot/assign_role.py
```diff
@@ -850,6 +850,16 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index, os_volume_size_gb,
               'path': '/properties/capabilities'}]
     ironic_client.node.update(node_uuid, patch)
 
+    # Detecting BOSS Card and find the boss device size
+    lst_ctrls = drac_client.list_raid_controllers()
+    boss_disk = [ctrl.id for ctrl in lst_ctrls if ctrl.model == "BOSS-S1"]
```
cdearborn: Can we change the above to something like cntrl.model.startswith("BOSS")?
This will make it work for versions other than just S1.

src/pilot/assign_role.py
```diff
@@ -850,6 +850,16 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index, os_volume_size_gb,
...
+    lst_physical_disks = drac_client.list_physical_disks()
```
cdearborn: I'm pretty sure there will be a problem when running assign_role against a node that does not have a BOSS card.  We need to detect that case and skip lines 856 through 862 in that case.

src/pilot/assign_role.py
```diff
@@ -850,6 +850,16 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index, os_volume_size_gb,
               'path': '/properties/capabilities'}]
     ironic_client.node.update(node_uuid, patch)
 
+    # Detecting BOSS Card and find the boss device size
```
cdearborn: We should only execute this code if the user has not specified os_volume_size on the command line.

src/pilot/assign_role.py
digambar15: > Preview
Sure Chis.

src/pilot/assign_role.py
digambar15: I have handled this scenario.

src/pilot/assign_role.py
digambar15: okay.

digambar15:
Completed all the changes as you suggested.

cdearborn:
Could we move the code changes inside of the select_os_volume() method, since the new code is all about selecting the BOSS card as the OS volume if it is present?

digambar15:
Sure Chris, I will move it under select_os_volume.

digambar15:
Let me fix that now and recheck again.

PR #37, opened 2018-03-29
Configuring logger to pull out urllib3 warnings

jproque-dell:
This is to get rid out the urllib3 warning we have since i changed all print statement to logger statement.
See below.
INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): 192.168.120.9

### DIFF
--- a/src/pilot/deploy-overcloud.py
+++ b/src/pilot/deploy-overcloud.py
@@ -664,6 +664,8 @@ def main():
                             help="Indicates if the deploy-overcloud script "
                                  "should be run in debug mode")
         args = parser.parse_args()
+
+        LoggingHelper.configure_logging()
         LoggingHelper.add_argument(parser)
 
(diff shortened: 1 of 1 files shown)


src/pilot/deploy-overcloud.py
```diff
@@ -664,6 +664,8 @@ def main():
                             help="Indicates if the deploy-overcloud script "
                                  "should be run in debug mode")
         args = parser.parse_args()
+
+        LoggingHelper.configure_logging()
```
uday344: This method requires arguments and tried adding this but it didn't work for me.
So I changed 
logging.basicConfig(level=logging.INFO)
to 
logging.basicConfig(level=logging.CRITICAL) 
in deploy-overcloud.py
and fixed it for now.

Do we require INFO messages as well ?

PR #38, opened 2018-03-30
Added vlock files for 10.2

cdearborn:
Added vlock files for 10.2


heydog15:
+1 Both files look good.  Compared to what I locked on both systems.

jtwill98:
looks good... merging

PR #40, opened 2018-03-30
Fix logging in deploy-overcloud.py

cdearborn:
This fixes:
- deploy-overcloud.py errors out due to bad configure_logging call
- urlib3 floods the log with logging messages


cdearborn:
I haven't tested this yet because I don't have the latest bits on my stamp.  Would like to merge this as-is since it can't break things worse then their already broken, and will test a reinstall from scratch.

jtwill98:
It can't be anymore broken than it is right now, so I'm merging it. Kurt said he will test.

PR #41, opened 2018-03-30
Fix update_ssh_config

cdearborn:


heydog15:
+1 Looks good to me. That's what I did on my system and it worked as expected.

PR #42, opened 2018-03-30
Updated to perform the steps required to install Dashboard and ceph …

jtwill98:
…collectors


src/pilot/config_dashboard.py
```diff
@@ -350,254 +293,304 @@ def prep_ceph_hosts(dashboard_node, ceph_nodes):
...
-    the Director adds an extra ".storage" subdomain to all Controllers'
-    storage network address. This causes the Calamari server to report an
-    FQDN for all monitors that don't match their actual FQDN, and that ends
-    up confusing the Storage Console.
+    home = expanduser("~")
+    root_key = Node.root_home + "/.ssh/authorized_keys"
+    tmp_key = "/tmp/authorized_keys"
+    repl_str = ".*sleep 10\" "
+    bak_date = dashboard_node.run('date +"%Y%m%d%H%M"')
+
```
pauldardeau: It might be helpful to add comment here indicating that it's a check for existence of file

src/pilot/config_dashboard.py
```diff
@@ -350,254 +293,304 @@ def prep_ceph_hosts(dashboard_node, ceph_nodes):
...
+                                                  && echo true \
+                                                  || echo false"
+                                                  .format(conf_file))
+            LOG.debug("STDOUT for node ({}) file ({}) = {}"
+                      .format(node.fqdn, conf_file, stdout))
+            if "true" in stdout:
+                node.run("mv {} {}.bak".format(conf_file, conf_file))
+                LOG.info("Restarting collectd service on node ({})"
+                         .format(node.fqdn))
+                node.run("systemctl restart collectd")
```
pauldardeau: Since this is within loop executing for each collectd conf file, it might be restarted repeatedly. Might be cleaner to set a boolean 'need_restart' and then restart outside of loop if set.

heydog15:
I see that there are 3 instances where Rhscon is defined in the script.  Should these b removed?

src/pilot/config_dashboard.py
```diff
@@ -493,6 +489,7 @@ def prep_collectd(dashboard_node, ceph_nodes):
     collectd_dir = "/etc/collectd.d"
 
     for node in ceph_nodes:
+        collectd_restart = "False"
```
pauldardeau: why use string for boolean?  why not False/True in place of "False"/"True"

src/pilot/config_dashboard.py
```diff
@@ -503,10 +500,13 @@ def prep_collectd(dashboard_node, ceph_nodes):
             LOG.debug("STDOUT for node ({}) file ({}) = {}"
                       .format(node.fqdn, conf_file, stdout))
             if "true" in stdout:
-                node.run("mv {} {}.bak".format(conf_file, conf_file))
-                LOG.info("Restarting collectd service on node ({})"
-                         .format(node.fqdn))
-                node.run("systemctl restart collectd")
+                node.run("sudo mv {} {}.bak".format(conf_file, conf_file))
+                collectd_restart = "True"
```
pauldardeau: see earlier comment about string vs boolean

src/pilot/config_dashboard.py
```diff
@@ -503,10 +500,13 @@ def prep_collectd(dashboard_node, ceph_nodes):
...
+
+        if "True" in collectd_restart:
```
pauldardeau: see earlier comment about string vs boolean

src/pilot/config_dashboard.py
```diff
@@ -328,15 +267,15 @@ def prep_ceph_hosts(dashboard_node, ceph_nodes):
         with open(tmp_hosts, "r") as f:
```
gaelrehault: I dont see tmp_hosts defined there ?

jtwill98: See line 263, tmp_hosts = os.path.join("/tmp", "hosts-{}".format(node.fqdn))

gaelrehault: thank you.. git hub collapsed/was hiding it

PR #43, opened 2018-04-02
Fixed XSP profile deployment failure

cdearborn:
The XSP profile was failing to deploy any compute nodes.  This turned out to be because DellComputeCount in dell-environment.yaml was not being updated.

This patch makes the deployer always update that count regardless of the profile being deployed.


dpaterson:
Fixes: CES-9485

PR #51, opened 2018-04-09
Fix deployment failure due to invalid characters from iDRAC

cdearborn:
This patch fixes an issue where on some stamps, when enumerating
DCIM_ControllerView, the iDRAC can return invalid unprintable
characters in the DriverVersion field.

This patch catches an XMLSyntaxError generated by etree.fromstring() during
enumeration and when it occurs, it strips out the unprintable characters and
reparses the XML in addition to logging the offending XML and the exception.


jtwill98:
Checked jenkins build failure, rebuilt the CI and looks OK ... not sure why Jenkins original build indicated a failure ... merging

PR #63, opened 2018-04-13
Performance Optimization

uday344:


src/deploy/osp_deployer/deployer.py
```diff
@@ -262,6 +276,45 @@ def deploy():
     logger.info("log : /auto_results/ ")
     sys.exit(ret_code)
 
+def validate_performance_optimization_params(
```
gaelrehault: Could you move that function into settings_sanity.py,   an add a call to  it in checkpoints.py :: verify_deployer_settings

src/deploy/osp_deployer/deployer.py
```diff
@@ -118,6 +118,10 @@ def deploy():
 
         logger.debug("loading settings files " + args.settings)
         settings = Settings(args.settings)
+        validate_performance_optimization_params(
```
gaelrehault: can be removed from here and incorporated in verify_deployer_settings, see comment below

src/deploy/osp_deployer/deployer.py
```diff
@@ -231,6 +235,16 @@ def deploy():
...
+        if settings.hpg_enable:
+            logger.info(
+                " HugePages has been successfully configured with size: " +
+                settings.hpg_size)
+        if settings.numa_enable:
+            logger.info(
+                " NUMA has been successfully configured"
+                " with hostos_cpus count: " +
+                settings.hostos_cpu_count)
+
```
gaelrehault: i beleive the the above is already in Directory.py :: summarize_deployment

src/deploy/osp_deployer/deployer.py
```diff
@@ -262,6 +276,45 @@ def deploy():
...
+    except Exception as error:
+        raise Exception(
+            "Please enter only integers for"
+            " innodb_buffer_pool_instances in .ini file.")
+    if innodb_buffer_pool_instances_int < 1 or \
+       innodb_buffer_pool_instances_int > 48:
+        logger.error(
+            "innodb_buffer_pool_instances not in range 1-48")
+        raise Exception("Invalid param innodb_buffer_pool_instances")
+
```
gaelrehault: I think we need to move those into the profiles where it would be best handled ,  create a is_int method  and an is in range one so those can be set in the .json file.

src/pilot/deploy-overcloud.py
```diff
@@ -636,7 +278,28 @@ def main():
         parser.add_argument("--hugepages_size",
                             dest="hugepages_size",
                             required=False,
+                            default="2MB",
                             help="HugePages size")
```
gaelrehault: default should be now be 1GIG

uday344: Fixed

src/deploy/osp_deployer/deployer.py
uday344: Methods added to profile.py
JSON updated

src/deploy/osp_deployer/deployer.py
uday344: Removed

src/deploy/osp_deployer/deployer.py
uday344: Validations moved to profile.py
Fixed

src/deploy/osp_deployer/deployer.py
uday344: Validations moved to profile.py
Fixed

gaelrehault:
Could you please rebase/resolve current conflicts ?

uday344:
All conflicts resolved

src/pilot/templates/dell-environment.yaml
```diff
@@ -33,7 +33,23 @@ parameter_defaults:
   # NovaVcpuPinSet: ['8-47']
  
   # Number of Dell Compute nodes 
-  DellComputeCount: 3
+
+  DellComputeCount: 0
```
cdearborn: Is there a compelling reason for changing DellComputeCount from 3 to 0?  I would prefer it to remain 3 since that is what our standard install uses.

src/pilot/templates/first-boot.yaml
```diff
@@ -14,13 +14,109 @@
...
+              echo "Applied _TunedProfile_ profile"
+            fi >> /tmp/tuned-applied.log 2>&1
+
+            if [[ $(hostname) == *controller* ]] ; then
+
+              # HAProxy Optimization
+              sed -i 's/\$haproxy_default_maxconn        =.*/\$haproxy_default_maxconn        = _HAProxyMaxconn_,/' /etc/puppet/modules/tripleo/manifests/haproxy.pp
+
+              # Openstack Service - Nova Optimization
+              sed -i 's/osapi_max_limit                      \=.*/osapi_max_limit                      = 10000,/' /etc/puppet/modules/nova/manifests/api.pp
```
cdearborn: I would prefer that we use ControllerExtraConfig to set these parameters when it is possible to use that method.  I know there are some parameters here where we can't do that, so we must use this method for those.

uday344:
Please note that I have resolved the first-boot conflict between DPDK and Performance Optimization. 
Now Performance Optimization will also work when DPDK is enable.

src/pilot/templates/neutron-ovs-dpdk.yaml
```diff
@@ -52,3 +52,25 @@ parameter_defaults:
   # Configure the classname of the firewall driver to use for implementing
   # security groups.
   NeutronOVSFirewallDriver: openvswitch
+
```
cdearborn: I don't think we need any of these changes here because all of this is already included in dell-environment.yaml, which is always included in the deploy command.

uday344: Removed parameter duplication.

PR #65, opened 2018-04-16
Profiles definition update

gaelrehault:


src/deploy/osp_deployer/profiles/xsp.json
```diff
@@ -81,11 +81,10 @@
                   "false"
                ]
             },
-            "use_ipmi_driver": "",
-            "use_in_band_introspection": ""
-
+         "use_ipmi_driver":"",
+         "use_in_band_introspection":"",
+         "enable_version_locking":""
```
cdearborn: It looks like the 3 lines above need to be indented.

cdearborn: oh, sorry, my mistake

PR #70, opened 2018-04-19
Changing dpdk_enable enforcement for CSP Profile

jproque-dell:
This fix allows the user to set dpdk to disable even if he's using a csp profile. It reduces enforcement but allows flexibility. Hopefully solving the bug Kurt went through.


heydog15:
One other question on this one.  In the .ini, it states that if ovs-dpdk is set to false, the user should comment out the lines for the interfaces.  Is that true?  I would think that setting it to false should ignore those lines even if they are uncommented out.

PR #73, opened 2018-04-20
updated default cSP profile settings

gaelrehault:


heydog15:
Do we want to comment out the lines for the ovs-dpdk interfaces since the help in the .ini file tells you to do so if you set the ovs-dpdk feature enablement to false?

gaelrehault:
Good point @heydog15 , thanks, fixed that.

PR #74, opened 2018-04-23
aligned sample ini's

gaelrehault:


heydog15:
Can we change the model numbers in both files to 14g?

gaelrehault:
we have another bug to change the server models

heydog15:
I know.  Just figured we might get em both knocked off in one PR.  Your call.

PR #78, opened 2018-04-25
OVS-DPDK Enhancement

uday344:


src/pilot/templates/post-install.yaml
```diff
@@ -1,60 +1,60 @@
...
+    type: OS::Heat::SoftwareConfig
+    properties:
+      group: script
+      config: |
+        #!/bin/bash
+
+        set -x
+        if [[ $HOSTNAME =~ "compute" ]] ; then
+          VCPUPINSET=$(crudini --get /root/parameters.ini DEFAULT novacpus)
+          crudini --set /etc/nova/nova.conf DEFAULT vcpu_pin_set $VCPUPINSET
```
jproque-dell: Why doing this? Is the NovaVcpuPinSet not already set to the appropriate values? If so, then the nova.conf will be already populated.

src/pilot/templates/post-install.yaml
```diff
@@ -1,60 +1,60 @@
...
-            grep -q "Before=.*" $tuned_service
-            if [ "$?" -eq 0 ]; then
-              sed -i 's/^\(Before=.*\)/\1 network.target openvswitch.service/g' $tuned_service
-            else
-              sed -i '/After/i Before=network.target openvswitch.service' $tuned_service
-            fi
-          fi
-          systemctl daemon-reload
-        fi
+# Copyright (c) 2016-2018 Dell Inc. or its subsidiaries.
```
jproque-dell: Why do we need a post-install script? What is its purpose?

src/pilot/templates/post-install.yaml
uday344: We can't set whole set of vcpus to in nova.conf as we need to subtract pmd cpus from vcpus and rest can go to nova.conf .
Refer -
[DEFAULT]
hostcpus = 0,24,2,26,1,25,3,27
pmdcpus = 5,29,4,28,6,30,8,32,7,31,9,33
novacpus = 10,34,12,36,14,38,16,40,18,42,20,44,22,46,11,35,13,37,15,39,17,41,19,43,21,45,23,47
isolcpus = 5,29,4,28,6,30,8,32,7,31,9,33,10,34,12,36,14,38,16,40,18,42,20,44,22,46,11,35,13,37,15,39,17,41,19,43,21,45,23,47
grub = default_hugepagesz=1GB hugepagesz=1G hugepages=116 iommu=pt intel_iommu=on

Please note that there are no changes to post-install.yaml (same as master's)

src/pilot/templates/post-install.yaml
uday344: To apply vcpus (after subtracting pmd cpus)  in nova.conf 
To add network.target and systemd-sysctl.service to tuned.service
This is related to tuned-adm

src/pilot/templates/post-install.yaml
jproque-dell: I see, but NovaVcpuPinSet is handled by the appropriate TripleO Paramater and therefore should managed anywhere else. So it's good to substract cpus, but once you have the remaining vcpus then you should update the line that goes along with the dell-environment.yaml:
# NovaVcpuPinSet: ['8-47']

src/pilot/templates/post-install.yaml
jproque-dell: see my comment below. Why do we need to set up these services after the deployment? This is part of the performance enhancement?

src/pilot/templates/post-install.yaml
uday344: We can apply NovaVcpuPinSet in environment and we are calculating pmdcpus in first-boot-ovs-dpdk.yaml so not possible to set that to NovaVcpuPinSet. 
We are applying the vcpus to NovaVcpuPinSet when dpdk is not enable.

src/pilot/templates/post-install.yaml
uday344: Note that this is not a part of performance optimization. 
This is related to tuned-adm cpu-partitioning profile which is required to set for dpdk.  
I think xlfow guys can explain this in details.

src/pilot/templates/post-install.yaml
uday344: https://github.com/uday344/JetPack/blob/3b13861d223ac8c1179f6834bb25d1996a10f0aa/src/pilot/dell_nfv.py#L273

jproque-dell:
Only thing i'm still concerned about, is what will happen if NUMA/Hugepages are enabled accidentally by the user and that the puppet manifest are ran afterward? Has it been tested?

gaelrehault:
Same concern as JP, what happens if ovs dpdk is enabled, and different values are set for numa/hp settings ?

jproque-dell:
It seems to me that this PR needs to be closed as its aim was to reorder the interaction of all the NFV features which is now part of another PR.

uday344:
Closing this PR due to https://github.com/dsp-jetpack/JetPack/pull/81

PR #80, opened 2018-04-27
CES-9514 - Changing reference from 13G to 14G servers.

jproque-dell:
This PR is about changing references from 13G to 14G servers.
We are expecting the interfaces to remain on the same slot than for 13G.


heydog15:
Do we have systems with that many NIC slots?  I don't see any in our lab that have 5 nic slots

PR #81, opened 2018-04-27
OVS-DPDK + NUMA + HUGEPAGES

uday344:


src/pilot/dell_nfv.py
```diff
@@ -55,50 +57,6 @@ def __init__(self, overcloud_name):
         self.overcloudrc = "source " + home_dir + "/"\
             + self.overcloud_name + "rc;"
 
-    def reboot_compute_nodes(self):
```
jproque-dell: I would have preferred to keep this one as it's easier to track and log the reboot. Besides, logging these reboot will help us to make sure it has been done successfully.

src/pilot/dell_nfv.py
```diff
@@ -211,62 +173,66 @@ def edit_environment_files(
...
-                str(enable_hugepage) +
-                '|" ' +
-                file_path)
-            cmds.append(
-                'sed -i "s|dellnfv::numa::enable:.*|dellnfv::numa::enable: ' +
-                str(enable_numa) +
-                '|" ' +
-                file_path)
-
+            if not ovs_dpdk:
```
jproque-dell: Why restricting this command to execute only if ovs_dpdk is enabled? Some of the filters included are needed to NUMA and Hugepages. NUMATopology and AggregateFilter are some of those.

src/pilot/dell_nfv.py
```diff
@@ -211,62 +173,66 @@ def edit_environment_files(
...
+                        "sed -i 's|NumaCpus:.*|NumaCpus: " +
+                        VCPUS +
+                        "|' " +
+                        file_path)
+                    cmds.append(
+                        'sed -i "s|  # NovaVcpuPinSet|  ' +
+                        'NovaVcpuPinSet|" ' +
+                        file_path)
+                    cmds.append(
+                        "sed -i 's|NovaVcpuPinSet:.*|NovaVcpuPinSet: \"" +
```
jproque-dell: I don't see this option within the dell-environment.yaml file. This should present and commented as default. If numa is enabled, then this command will succeed.

src/pilot/templates/first-boot-ovs-dpdk.yaml
```diff
@@ -426,36 +352,43 @@ resources:
...
+      config:
+        str_replace:
+          template: |
+            #!/bin/bash
+            if [[ $(hostname) == *compute* ]] ; then
+              sudo grubby --update-kernel=ALL --args="_HugePages_"
+              echo "HugePages are applied"
+              sudo grubby --update-kernel=ALL --args="isolcpus=_NumaCpus_"
+              echo "Numa is applied"
+              reboot
```
jproque-dell: imho, this reboot needs to be pulled out and use the function for post-deploy tasks as mentioned earlier.

src/pilot/templates/first-boot.yaml
```diff
@@ -43,6 +69,37 @@ resources:
...
+                sudo grubby --update-kernel=ALL --args="_HugePages_"
+                echo "HugePages are applied"
+              fi
+              numaflag=_NumaEnable_
+              if [[ "$numaflag" == "True" ]] ; then
+                sudo grubby --update-kernel=ALL --args="isolcpus=_NumaCpus_"
+                echo "Numa is applied"
+              fi
+              if [[ "$hpgflag" == "True" || "$numaflag" == "True" ]] ; then
+                echo "Rebooting ..."
```
jproque-dell: Same.

src/pilot/dell_nfv.py
uday344: I am not sure about it.
In master code we rebooting twice in case dpdk is enable. (In first-boot.yaml and Post deployment)
If we remove reboot from first-boot-ovs-dpdk.yaml then dpdk won't work as NUMA and Hugepages are prerequisites for DPDK.
We did same mistake earlier by removing NUMA HUGPAGES code from first-boot-ovs-dpdk.yaml and it didn't work.
So we have to keep the reboot and also we don't want to reboot twice.

src/pilot/dell_nfv.py
uday344: It is there 
https://github.com/uday344/JetPack/blob/cdf64a095075b864073bf5a686b78fc01bbd81b5/src/pilot/templates/dell-environment.yaml#L30

src/pilot/dell_nfv.py
jproque-dell: yes i just figure that out. This reboot is done after deployment as opposed to your reboot which is done after the first boot but before Puppet does his job. Hence your failure.

src/pilot/dell_nfv.py
jproque-dell: So but if none of the features are enabled, ie neither Hugepages, nor numa nor dpdk then default filters apply.

src/pilot/dell_nfv.py
jproque-dell: gotcha.

src/pilot/templates/first-boot-ovs-dpdk.yaml
jproque-dell: Understood.

src/pilot/templates/first-boot.yaml
jproque-dell: Agree.

src/pilot/dell_nfv.py
uday344: Yes, This is same as way to old master's code 
https://github.com/uday344/JetPack/blob/ebaa424ce77b04edc1888dd71612f55283be1539/src/pilot/deploy-overcloud.py#L216

PR #83, opened 2018-05-03
CES-9561 & CES-9562 FIX

uday344:

### DIFF
--- a/src/pilot/cpu_siblings.py
+++ b/src/pilot/cpu_siblings.py
@@ -19,9 +19,9 @@
 # {'number_of_host_os_cpu' : {
 # 'mode2_rem_cores': 'Remaining cores after removing mode2_pmd_cores',
 # 'mode1_rem_cores': 'Remaining cores after removing mode1_pmd_cores',
+# 'vcpu_pin_set' : 'Range of corresponding CPUs available for nova',
 # 'mode1_pmd_cores': 'Range of corresponding CPUs for pmd in Mode 1',
 # 'mode2_pmd_cores': 'Range of corresponding CPUs for pmd in Mode 2',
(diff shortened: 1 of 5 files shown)


src/pilot/cpu_siblings.py
```diff
@@ -16,45 +16,221 @@
 
 
 # It is a dictionary of the form {'total number of cpu cores' :
-# {'number_of_host_os_cpu' : {'vcpu_pin_set' : 'Range of corresponding cpus',
-# 'host_os_cpu' : 'Range of corresponding cpus'} } }
+# {'number_of_host_os_cpu' : {
```
jproque-dell: Can you make the description using the same order you have coded up below? It will be easier to read .

src/pilot/cpu_siblings.py
```diff
@@ -16,45 +16,221 @@
...
+# 'vcpu_pin_set' : 'Range of corresponding cpus',
```
jproque-dell: I would prefer "Range of corresponding cpus available for nova"

src/pilot/cpu_siblings.py
```diff
@@ -16,45 +16,221 @@
...
+# 'host_os_cpu' : 'Range of corresponding cpus',
```
jproque-dell: I would prefer "Range of corresponding cpus available for Host OS"

src/pilot/cpu_siblings.py
uday344: Fixed

src/pilot/cpu_siblings.py
uday344: Fixed

src/pilot/cpu_siblings.py
uday344: Fixed

PR #84, opened 2018-05-04
CES-9058 -- Updated code and templates with latest supported RHEL ... 

jtwill98:
…release (7.5)


gaelrehault:
rhel7 doesn't seem to be a valid os-variant , we probably should be using "rhel7.5"

see http://libguestfs.org/virt-builder.1.html 
--os-variant is highly recommended, because it will present optimum devices to enable the guest to run most efficiently. To get a list of all variants, do:
 osinfo-query os

PR #87, opened 2018-05-09
Implement configure_raid support for boss card

digambar15:
1. Implement configure raid for boss card in assign_role.py
2. Find out boss controller for detecting physical disks
3. Added code in dracclient for configuring raid for boss card
4. Updated dracclient_raid.patch


src/pilot/assign_role.py
```diff
@@ -267,15 +267,24 @@ def define_target_raid_config(role, drac_client):
...
+    if fqdd.startswith('RAID.'):
+        return True
+    elif fqdd.startswith('BOSS'):
+        return True
+    else:
+	disk_controllers = drac_client.list_raid_controllers()
+        boss_raid_controller = [
+            c.id for c in disk_controllers if c.model.startswith('BOSS')]
+        if fqdd == boss_raid_controller[0]:
+            return True
```
cdearborn: Is this else clause needed?  Isn't the check to see if the FQDD starts with BOSS above enough?

src/pilot/dracclient_raid.patch
```diff
@@ -125,7 +125,19 @@
...
          """Converts a list of physical disks into or out of RAID mode.
-@@ -396,13 +428,19 @@
+@@ -383,6 +415,11 @@
+             utils.validate_integer_value(size_mb, 'size_mb', error_msgs)
+ 
+         virtual_disk_prop_names.append('Size')
++
++        # Boss controller validation
++        if raid_controller.startswith('AHCI.Slot'):
++            size_mb = 0
```
cdearborn: I would really prefer to keep all of the BOSS specific logic in our downstream code, especially since we already have a method in assign_role to determine if the controller is a BOSS card.  Specifically, I think this logic should be in assign_role.

src/pilot/assign_role.py
```diff
@@ -399,8 +420,14 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
     # Define a logical disk to host the operating system.
-    os_logical_disk = define_storage_operating_system_logical_disk(
-        raid_cntlr_physical_disks, raid_controller_name)
+    raid_ctrl_name = raid_controller_name
+    raid_disks = raid_cntlr_physical_disks
+    boss_dsk_info = is_boss_controller(raid_disks, raid_ctrl_name, drac_client)
+    if boss_dsk_info:
+        os_logical_disk = boss_dsk_info
+    else:
+        os_logical_disk = define_storage_operating_system_logical_disk(
```
cdearborn: Let's remove the above added lines and just leave this as a call to define_storage_operating_system_logical_disk()

src/pilot/assign_role.py
```diff
@@ -373,6 +382,18 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
         key=physical_disk_id_to_key)
 
 
+def is_boss_controller(raid_disks, raid_ctrl_name, drac):
+    if is_raid_controller(raid_ctrl_name, drac):
```
cdearborn: This method should just return True if the controller is a BOSS and False if not.

src/pilot/assign_role.py
```diff
@@ -267,15 +267,24 @@ def define_target_raid_config(role, drac_client):
         'logical_disks': logical_disks} if logical_disks is not None else None
 
 
-def is_raid_controller(fqdd):
-    return fqdd.startswith('RAID.')
+def is_raid_controller(fqdd, drac_client):
+    if fqdd.startswith('RAID.'):
+        return True
+    elif fqdd.startswith('BOSS'):
```
cdearborn: This method should just return is_boss_controller(fqdd), and the is_boss_controller() method should contain the logic that detects the BOSS controller.

src/pilot/assign_role.py
digambar15: ok. got it.

src/pilot/dracclient_raid.patch
digambar15: ok. do you mean we dont need to add code here in dracclient_raid ?
I have added here because it will be difficult to set size to 0 if boss controller is detected and if I set size=0 from assign_role, execution fails. 
I will retest by setting size=0 in define_storage_operating_system_logical_disk() method.

src/pilot/assign_role.py
digambar15: Got it.

src/pilot/assign_role.py
digambar15: okay

src/pilot/assign_role.py
digambar15: okay

src/pilot/dracclient_raid.patch
cdearborn: Yes, I would prefer not to add code here.  I would think we could detect the BOSS and set the size to 0 in define_storage_operating_system_logical_disk and that should work...

src/pilot/assign_role.py
digambar15: To skip the call, define_storage_operating_system_logical_disk() we added above code. 
Reason to skip that is - below lines in the method - 
os_logical_disk_size_gb,
     os_physical_disk_names) = find_physical_disks_for_storage_os(
        physical_disks)

src/pilot/dracclient_raid.patch
digambar15: I will try again

digambar15:
I changed PR target branch from master to experimental_configure_raid-boss_card.

PR #90, opened 2018-05-10
DPDK 5 port fix for bug 9576

HNMN3:
- Arranged OVS bridges and interfaces in proper order.
 - Fixed Linux bonds and interfaces.


gaelrehault:
Please add the missing header on the files.

PR #91, opened 2018-05-10
CES-9577 - fixed compute nodes getting DHCP i/o static

gaelrehault:
fixed compute nodes getting DHCP address when static ip's were set in .properties

### DIFF
--- a/src/pilot/templates/static-ip-environment.yaml
+++ b/src/pilot/templates/static-ip-environment.yaml
@@ -25,6 +25,13 @@ resource_registry:
   OS::TripleO::Compute::Ports::StorageMgmtPort: ./overcloud/network/ports/noop.yaml
   OS::TripleO::Compute::Ports::TenantPort: ./overcloud/network/ports/tenant_from_pool.yaml
 
+
+  OS::TripleO::DellCompute::Ports::ExternalPort: ./overcloud/network/ports/noop.yaml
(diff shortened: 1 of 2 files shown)


src/pilot/templates/static-ip-environment.yaml
```diff
@@ -25,6 +25,13 @@ resource_registry:
   OS::TripleO::Compute::Ports::StorageMgmtPort: ./overcloud/network/ports/noop.yaml
   OS::TripleO::Compute::Ports::TenantPort: ./overcloud/network/ports/tenant_from_pool.yaml
 
+
```
cdearborn: I think it will be confusing to leave the OS::TripleO::Compute entries in this file.  Can you remove them?  I think it should be safe to do.

src/pilot/templates/static-ip-environment.yaml
```diff
@@ -84,6 +91,26 @@ parameter_defaults:
     - 192.168.170.31
     - 192.168.170.32
     - 192.168.170.33
+
```
cdearborn: I think it will be confusing to leave the NovaComputeIPs: entries in this file.  Can you remove them?  I think it should be safe to do.

gaelrehault:
@cdearborn the overcloud fails to deploy if i get rid of those Compute references

PR #92, opened 2018-05-11
DPDK 4 port fix for bug 9576

HNMN3:
- Fixed bug 9576 for OVS-DPDK 4 Port.
 - Arranged Bridges, and Interfaces in Proper order.

### DIFF
--- a/src/pilot/templates/nic-configs/dpdk_mode1_4_port/compute-ovs-dpdk-mode1.yaml
+++ b/src/pilot/templates/nic-configs/dpdk_mode1_4_port/compute-ovs-dpdk-mode1.yaml
@@ -141,52 +141,37 @@ resources:
         os_net_config:
           network_config:
             -
-              type: ovs_bridge
-              name: br-tenant
-              use_dhcp: false
-              dns_servers: {get_param: DnsServers}
-              addresses:
-                -
-                  ip_netmask:
-                    list_join:
-                      - '/'
-                      - - {get_param: ControlPlaneIp}
-                        - {get_param: ControlPlaneSubnetCidr}
-              routes:
-                -
-                  ip_netmask: 169.254.169.254/32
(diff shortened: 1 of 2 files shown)


src/pilot/templates/nic-configs/dpdk_mode1_4_port/compute-ovs-dpdk-mode1.yaml
```diff
@@ -141,52 +141,37 @@ resources:
...
-              addresses:
+              type: linux_bond
+              name: bond0
+              bonding_options: {get_param: ComputeBondInterfaceOptions}
+              members:
                 -
-                  ip_netmask:
-                    list_join:
-                      - '/'
-                      - - {get_param: ControlPlaneIp}
```
cdearborn: In the resulting file, nothing is assigned the control plane IP.  I would think this would result in no interface being on the provisioning network, and we have to have an interface on the provisioning network.

src/pilot/templates/nic-configs/dpdk_mode2_4_port/compute-ovs-dpdk-mode2.yaml
```diff
@@ -141,52 +141,37 @@ resources:
...
-              addresses:
+              type: linux_bond
+              name: bond0
+              bonding_options: {get_param: ComputeBondInterfaceOptions}
+              members:
                 -
-                  ip_netmask:
-                    list_join:
-                      - '/'
-                      - - {get_param: ControlPlaneIp}
```
cdearborn: In the resulting file, nothing is assigned the control plane IP. I would think that provisioning will fail on the node because it's not on the provisioning network.

cdearborn:
Closing this PR in favor of https://github.com/dsp-jetpack/JetPack/pull/95.
We discussed this and agreed that DPDK would not be supported with the 4_port NIC configuration.

PR #97, opened 2018-05-15
Fix CES-9604 values for range

digambar15:
Fixing value range in is_in_range() method which causes actual failure in if statement.

### DIFF
--- a/src/deploy/osp_deployer/profile.py
+++ b/src/deploy/osp_deployer/profile.py
@@ -151,10 +151,10 @@ class Profile():
 
     def is_in_range(self, value, input):
         valid = True
-        list = input.split('-')
-        min = int(list[0])
-        max = int(list[1])
-        if int(value) not in range(min, max):
+        val_list = input.split('-')
+        min_val = int(val_list[0])
+        max_val = int(val_list[1])
+        if int(value) not in range(min_val, max_val+1):
             valid = False
         return valid
 


jproque-dell:
Would rather to have:
def is_in_range(self, value, input):
        valid = True
        list = input.split('-')
        min = int(list[0])
        max = int(list[1]) + 1
        if int(value) not in range(min, max):
            valid = False
        return valid

src/deploy/osp_deployer/profile.py
```diff
@@ -151,10 +151,10 @@ def is_int(self, value):
 
     def is_in_range(self, value, input):
         valid = True
-        list = input.split('-')
-        min = int(list[0])
-        max = int(list[1])
-        if int(value) not in range(min, max):
+        val_list = input.split('-')
```
jproque-dell: Why do we need to change the variable name?

src/deploy/osp_deployer/profile.py
```diff
@@ -151,10 +151,10 @@ def is_int(self, value):
...
+        min_val = int(val_list[0])
```
jproque-dell: Same.

src/deploy/osp_deployer/profile.py
```diff
@@ -151,10 +151,10 @@ def is_int(self, value):
...
+        max_val = int(val_list[1]) + 1
```
jproque-dell: Same.

src/deploy/osp_deployer/profile.py
digambar15: Updated PR as per changes suggested

digambar15:
Updated PR as per the changes suggested

PR #98, opened 2018-05-16
Change min value for maraidb_max_connection and innodb_buffer_pool_instances

digambar15:
- Added min value for mariadb_max_connection to 151 which is default in mariadb
- Changes min value for innodb_buffer_pool_instances to 8 which is also default


suryaprabhakar:
make sure the comment indicates that we are setting minimum values to system default

digambar15:
Thank you Chris!

PR #99, opened 2018-05-21
ces-9624: static-ips.yaml has both DellCompute and NovaCompute sections

cdearborn:
Removed the CHANGEME's and sample IPs from the NovaCompute section of static-ip-environment.yaml.

Updated the automation code so it won't substitute IPs into the NovaCompute section.


heydog15:
+1

PR #104, opened 2018-05-25
Replaced grubby command with sed command for HugePages and NUMA deployment.

jainprasuk1996:


alp317:
Same changes are also required in the first-boot-ovs-dpdk.yaml file.

PR #105, opened 2018-05-25
Patch roles_data.yaml and current_config_hosts.rb

dpaterson:
Both of these patches fix update, otherwise
upate will fail.

Fixes: CES-9644, CES-9630


heydog15:
+ 1. Looks good, at least for the Manila issue.  Not sure on the other one.

PR #106, opened 2018-05-25
Updated vlock files to include Red Hat fixes for latest Spectre and …

jtwill98:
…Meltdown side-channel vulnerabilities.


jtwill98:
I wasn't on the standup today, but according to Kurt once we took the latest bits and was able to at minimal verify the bits installed and passed sanity tests on a 14G referenced architecture, we were to submit new versionlock files.  Here there are.  Looks like a lot of changes but the is because the file ordering changed.

heydog15:
+1

PR #110, opened 2018-06-13
Add support for DSS9000

cdearborn:
- This patch adds support for the DSS9000 by adding a NIC entry to
  dell_systems.json.
- It also adds support for storage nodes that only have an HBA330 and no BOSS
  card since this is the way that the DSS9000 storage nodes were configured.

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -881,14 +881,18 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
     controllers = drac_client.list_raid_controllers()
 
     found_hba = False
+    found_boss = False
     for controller in controllers:
         if "hba330" in controller.model.lower():
             found_hba = True
-            break
+        if "boss" in controller.model.lower():
+            found_boss = True
 
-    if not found_hba:
-        LOG.info("No HBA330 found.  Not generating OSD config for "
-                 "{ip}".format(ip=ip_mac_service_tag))
+    if not (found_hba and found_boss):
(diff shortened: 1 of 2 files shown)


heydog15:
Looks good.

src/pilot/assign_role.py
```diff
@@ -881,13 +881,15 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
         if "hba330" in controller.model.lower():
             found_hba = True
-            break
+        if "boss" in controller.model.lower():
+            found_boss = True
 
-    if not found_hba:
-        LOG.info("No HBA330 found.  Not generating OSD config for "
+    if not (found_hba and found_boss):
+        LOG.info("Both BOSS and HBA330 are not present.  Not generating OSD config for "
```
pauldardeau: Log message wording is awkward.  Suggest "Neither BOSS nor HBA330 are present" to make more clear.

gaelrehault: agreed

cdearborn: The intent is that BOTH the boss card and the hba must be present for us to generate the OSD config.  Your wording makes it sound like we would generate the config if either one is present.

pauldardeau: Ok, my mistake. Agreed with your original wording then!

cdearborn:
I took another shot at the wording of the error message.  It's longer now, but I think it is more clear.

heydog15:
Looks better to me.

PR #123, opened 2018-07-24
fixes for sanity_test.sh (CLI) and enable_repos.py (repo updates)

jtwill98:


heydog15:
+1 This is the code base that I have tested in both deploying and cleaning up sanity networks.

PR #126, opened 2018-07-26
Support for 13G OSD configs on 13.0 

alp317:


gaelrehault:
I managed to deploy with 13g storage nodes set with CephPoolDefaultSize = 1
Can we calculate the settings to apply based on the number of OSD's available dynamically as part of those changes @alp317 ?

PR #134, opened 2018-08-07
re enabling overcloud image customization

gaelrehault:


heydog15:
I uncommented the first echo as well but looks good.  Merge it! :-)

PR #140, opened 2018-08-13
Only require HBA330 to be present for auto OSD config generation

cdearborn:
The DSS9k nodes did not have BOSS cards, only HBA330s.  This change makes it so that only an HBA330 needs to be present for assign_role to auto generate the Ceph OSD configuration.


heydog15:
Looks good.  This is what I tested on the DSS

PR #142, opened 2018-08-14
Deleted a print statement with a misspelling

cdearborn:
Deleted a print statement with a misspelling that snuck in somehow.


heydog15:
Looks good!!

PR #143, opened 2018-08-16
Refix creating a RAID10 with 16+ drives

cdearborn:
This patch adds back in a patch to the DRAC driver in Ironic so that RAID10s with 16 or more drives can be created.


heydog15:
Looks good.

PR #151, opened 2018-08-19
CES-10038 updated sample configs

gaelrehault:


heydog15:
In the PR merged for vlan aware, they also added the following:

# Note Additional one instance would be created for vlan testing. Keep the
# maxinmum instances to 9.

This needs to be added plus reworded.  I would put:

# Note: One additional instance will be created for vlan aware testing. Maximum
# number of instances can be 9 unless the quota is increased.

gaelrehault:
Closing this, adressed in https://github.com/dsp-jetpack/JetPack/pull/152

PR #153, opened 2018-08-21
Fix for CES-10049: Deployment fails with OSD configuration error

cdearborn:
This adds back in the code that checks to see if the generated OSD config is
different from the current OSD config.  This prevents erroneous failures when
redeploying just the overcloud.

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -925,25 +925,35 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
 
     if system_id in node_data_lookup:
         current_osd_config = node_data_lookup[system_id]
-        generated_config = json.dumps(new_osd_config, sort_keys=True,
-                                      indent=2, separators=(',', ': '))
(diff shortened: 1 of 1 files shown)


src/pilot/assign_role.py
```diff
@@ -925,25 +925,35 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
+            generated_config = json.dumps(new_osd_config, sort_keys=True,
+                                          indent=2, separators=(',', ': '))
+            current_config = json.dumps(current_osd_config, sort_keys=True,
+                                        indent=2, separators=(',', ': '))
+            raise RuntimeError("The generated OSD configuration for "
+                               "{ip_mac_service_tag} ({system_id}) is "
+                               "different from the one in {osd_config_file}.\n"
+                               "Generated:\n{generated_config}\n\n"
+                               "Current:\n{current_config}\n\n"
+                               "If this is unexpected then check for failed "
```
heydog15: Add a , after unexpected

heydog15:
+1

PR #154, opened 2018-08-22
CES-9803 Configure baremetal server's boot mode to UEFI or BIOS

digambar15:
1. Add support for UEFI in bood mode settings
2. Changed config_legacy_boot_mode to config_boot_mode
3. Implement the logic to support UEFI
4. Moved restart idrac code to schedule_restart_job() method

### DIFF
--- a/src/pilot/config_idrac.py
+++ b/src/pilot/config_idrac.py
@@ -19,6 +19,7 @@ import logging
 import os
 import requests.packages
 import sys
+
 from arg_helper import ArgHelper
 from constants import Constants
 from credential_helper import CredentialHelper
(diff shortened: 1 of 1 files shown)


digambar15:
[ BootMode(id='IPL', name='BootSeq', is_current=False, is_next=False), 
  BootMode(id='BCV', name='HddSeq', is_current=False, is_next=False), 
  BootMode(id='UEFI', name='UefiBootSeq', is_current=True, is_next=True), 
  BootMode(id='OneTime', name='OneTimeBootMode', is_current=False, is_next=False), 
  BootMode(id='vFlash', name='vFlash Boot Configuration', is_current=False, is_next=False)
]

As you can see UEFI is set as current boot mode.
but Bios sanity test is still pending.

Issue #156, opened 2018-08-28
ImportError: cannot import name hosts


goneri:
As discussed with @gaelrehault:

```
Traceback (most recent call last): 
 File "./deploy-overcloud.py", line 30, in <module> 
   from novaclient.v2 import hosts 
ImportError: cannot import name hosts
```

deploy-overcloud.py fails with the following error. `hosts` is not in `novaclient.v2 anymore`. The `hosts` module is not used by `deploy-overcloud.py` anyway. I applied the following change and restarted the script:

```diff
diff --git a/src/pilot/dell_nfv.py b/src/pilot/dell_nfv.py
index 318f062..95af9d2 100644
--- a/src/pilot/dell_nfv.py
+++ b/src/pilot/dell_nfv.py
@@ -26,7 +26,6 @@ import logging
 import requests
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
 from novaclient.v2 import aggregates
-from novaclient.v2 import hosts
 from novaclient.v2 import servers
 from credential_helper import CredentialHelper
 from datetime import datetime
diff --git a/src/pilot/deploy-overcloud.py b/src/pilot/deploy-overcloud.py
index 04e1738..ab2f61e 100755
--- a/src/pilot/deploy-overcloud.py
+++ b/src/pilot/deploy-overcloud.py
@@ -27,7 +27,6 @@ import novaclient.client as nova_client
 import time
 from command_helper import Ssh
 from novaclient.v2 import aggregates
-from novaclient.v2 import hosts
 from novaclient.v2 import servers
 from ironic_helper import IronicHelper
 from logging_helper import LoggingHelper
```

`get_dell_compute_nodes_hostnames()` still uses the `hosts` module. python-openstackclient provides a similar method: https://github.com/openstack/python-openstackclient/blob/master/openstackclient/compute/v2/host.py#L24

goneri:
`v2.hosts` was removed in Queens: https://github.com/openstack/python-novaclient/commit/1d88717e751e743ca7edd978346b95b6e7b3f235#diff-4ef50ccc65413eb1dda481e0484e831b

goneri:
`python-novaclient-9.1.1-1.el7ost.noarch` was initially released with OSP13 because of https://review.openstack.org/#/c/571540/ . The last puddle now ships the 10 package, this is the reason why @gaelrehault didn't notice the issue.

gaelrehault:
Should be fixed now @goneri

PR #158, opened 2018-08-30
removed references to removed nova module

gaelrehault:
addressing #156


jtwill98:
It looks syntactically sound and Kurt says it's been tested so I am merging it.

PR #161, opened 2018-09-06
CES-9803 - Configure baremetal server's boot mode to UEFI or BIOS

digambar15:


rpioso:
@cdearborn, thank you for your review. In response to your comments, the original intention of `reboot_required` was retained so that the processing is correct for a system whose boot mode is not flexibly programmable.

When that is the case, the boot mode is set to BIOS. Subsequent processing must respect the need to reboot or not for that. As was the case prior to this proposed change, only (1) reboot job is created and all of the jobs are scheduled to execute together.

PR #166, opened 2018-09-21
CES-9802 - Configure NICs' PXE boot settings for UEFI

digambar15:


src/pilot/config_idrac.py
```diff
@@ -110,10 +110,51 @@ def configure_nics_boot_settings(
...
+               "PxeDev3EnDis": "Disabled",
+               "PxeDev4EnDis": "Disabled",
+               "PxeDev1Interface": pxe_nic_id,
+               "PxeDev1Protocol": "IPv4",
+               "PxeDev1VlanEnDis": "Disabled"
+               }
+
+            provisioning_mac = nic.mac_address.lower()
+            response = drac_client.set_bios_settings(settings)
+            reboot_required = response['commit_required']
```
cdearborn: The above line uses the 'commit_required' attribute, which has been deprecated.  It should use the 'is_reboot_required' attribute instead.  See: https://github.com/openstack/python-dracclient/blob/master/dracclient/utils.py#L156

src/pilot/config_idrac.py
```diff
@@ -110,10 +110,51 @@ def configure_nics_boot_settings(
...
+
+            if response['commit_required']:
```
cdearborn: The above line uses the 'commit_required' attribute, which has been deprecated.  It should use the 'is_commit_required' attribute instead.

src/pilot/config_idrac.py
cdearborn: After discussing the above with Rick, given that this is being developed on a 10.2 cluster, I'm ok with leaving the above as-is for now.

src/pilot/config_idrac.py
cdearborn: After discussing the above with Rick, given that this is being developed on a 10.2 cluster, I'm ok with leaving the above as-is for now.

digambar15:
will fix those attributes in future release

PR #172, opened 2018-09-28
added BOSS support patches

gaelrehault:


src/pilot/install-director.sh
```diff
@@ -283,6 +283,17 @@ echo "## Patching ironic.conf..."
 apply_patch "sudo patch -b -s /etc/ironic/ironic.conf ${HOME}/pilot/ironic.patch"
 echo "## Done."
 
+# These patches add support for BOSS cards
+echo 
+echo "### Patching raid_conig_schema"
```
cdearborn: Can you change the above to "raid_config_schema"?

PR #180, opened 2018-10-11
Check added for CSP and XSP profiles to validate the nic_env_file values

beenum22:
A check has been added for CSP and XSP profiles to validate the nic config files:

- 4_port/nic_environment.yaml
- 5_port/nic_environment.yaml
- sriov_7_port/nic_environment.yaml
- sriov_9_port/nic_environment.yaml
- ovs-dpdk_7_port/nic_environment.yaml
- ovs-dpdk_9_port/nic_environment.yaml
- dvr_7_port/nic_environment.yaml
- ovs-dpdk_sriov_9_port/nic_environment.yaml

For example:
When the user sets 'sriov_enable" to "true" but doesn't uncomment the required "ComputeSriovInterfaceX" parameters and selects a wrong "nic_env_file". This check catches such cases and makes sure correct nic environment files are used.


src/deploy/osp_deployer/profile.py
```diff
@@ -168,3 +175,23 @@ def valid_buffer_pool_size(self, value):
...
+                    config_file == "ovs-dpdk_9_port/nic_environment.yaml"
+            elif self.settings.enable_dvr:
+                assert config_file == "5_port/nic_environment.yaml" or \
+                    config_file == "dvr_7_port/nic_environment.yaml"    
+            else:
+                assert config_file == "5_port/nic_environment.yaml" or \
+                    config_file == "4_port/nic_environment.yaml"
+            return True
+        except AssertionError:
+            return False
```
gaelrehault: This sort of defeats the intent of the nic env file implementation ..i think i 'd happy with just having a list of allowed nic's files in the profile definition and not the above.
We're trying to keep all the profile related information outside of the codes and contained within the profile definition. The above would require maintenance every time we add/remove nic files.
@cdearborn any opinion ?

cdearborn: I agree.  We should not be hard coding the names of the NIC env files.  This would force code changes every time we wanted to add or remove NIC env files.

gaelrehault:
@beenum22  could you update as per the comments & fix conflict  please ?

beenum22:
@gaelrehault I have removed the check and resolved the conflicts.

PR #182, opened 2018-10-11
CES-10143 fix

beenum22:
OVS-DPDK driver type "vfio-pci" requires intel iommu to be enabled in the grub. It was only being enabled for SR-IOV previously. It has been enabled in the kernel arguments as default now.


gaelrehault:
This wont affect  deployments where we have no ovs dpdk nor sriov ?

beenum22:
Nope. It should not affect the normal deployments. However, just to be on the safe side, I'll change to be enabled for OVS-DPDK and SR-IOV only.

PR #187, opened 2018-10-25
Fix for RAID creation bug

cdearborn:
This patch provides a work around for the RAID creation bug.

It works by creating the RAID volumes using realtime mode if the RAID controller supports this mode.  This allows the volume to be created without rebooting the server.

Tested with PERC H730P and PERC H740P with the latest firmware.


src/pilot/install-director.sh
```diff
@@ -261,8 +268,16 @@ apply_patch "sudo patch -b -s /usr/lib/python2.7/site-packages/dracclient/wsman.
 sudo rm -f /usr/lib/python2.7/site-packages/dracclient/wsman.pyc
 sudo rm -f /usr/lib/python2.7/site-packages/dracclient/wsman.pyo
 
+# This hacks in a patch to enable realtime RAID creation.
+echo
+echo "## Patching Ironic iDRAC driver job.py..."
+apply_patch "sudo patch -b -s /usr/lib/python2.7/site-packages/dracclient/resources/job.py ${HOME}/pilot/job.patch"
+sudo rm -f /usr/lib/python2.7/site-packages/dracclient/job.pyc
+sudo rm -f /usr/lib/python2.7/site-packages/dracclient/job.pyo
```
rpioso: I believe the path to the files being removed by the two (2) lines above is incorrect. Shouldn't the path be `/usr/lib/python2.7/site-packages/dracclient/resources/` ?

src/pilot/raid.patch
```diff
@@ -17,3 +45,51 @@
...
+                 physical_disks=logical_disk['physical_disks'],
+@@ -742,7 +753,10 @@
+                 span_length=logical_disk.get('span_length'),
+                 span_depth=logical_disk.get('span_depth'))
+ 
+-        return _commit_to_controllers(node, list(controllers))
++            if result['is_reboot_required'] == reboot_is_required:
++                realtime = False
++
++        return _commit_to_controllers(node, list(controllers), realtime)
```
rpioso: If I understand correctly, this would not correctly work when a node has multiple RAID controllers with differing real time capabilities. For example:

* Two (2) RAID controllers
* One has real time, so it returns `result['is_reboot_required']` set to `drac_constants.RebootRequired.optional`
* One does not have real time, so it returns `result['is_reboot_required']` set to `drac_constants.RebootRequired.true`

At the end of the loop, the variable `realtime` will be set to `False`, and `_commit_to_controllers()` will not use real time at all. The RAID controller that has real time time would still encounter the bug this PR is attempting to fix.

cdearborn: You are making huge assumptions on how the iDRAC will behave when a server has multiple RAID controllers, some of which support realtime mode and some of which don't.  Before we start writing code to handle that scenario, we would need to test the behavior of hardware in that configuration.  I'm not even sure that such a configuration is valid.
Also, JetPack doesn't support having multiple RAID controllers in a server, so your point is moot.

rpioso: A Dell EMC RAID and WS-Man subject matter expert confirmed that an iDRAC can manage multiple RAID controllers, some of which support real-time mode and some of which don't. And, as the ironic iDRAC driver code in question implements, virtual volumes can be created on them at the same time.

An example is a server that has two (2) RAID controllers, a PERC and BOSS. Recent generations of PERC support real-time, however, BOSS does not. Real-time and reboot should be specified when creating the two (2) configuration jobs, respectively. The iDRAC executes the real-time job for the PERC first. After that job completes, it reboots the system and executes the staged job for the BOSS.

I understand that configuration may not be supported by JetPack, in which case, this downstream patch is fine, as is. However, please note this change should not be pushed to upstream ironic without testing against such a configuration.

PR #190, opened 2018-10-30
Added the comments for dvr_7_port interfaces in sample .ini files.

rashidxf:


src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -228,6 +228,12 @@ StorageBondInterfaceOptions=mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 la
 #ComputeOvsDpdkInterface2=p3p1
 #BondInterfaceOvsOptions=bond_mode=balance-tcp lacp=active
 
+# To enable DVR with separate bond for Floating network on compute nodes, two interfaces should be used.
```
heydog15: I would add "a" before separate.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -228,6 +228,12 @@ StorageBondInterfaceOptions=mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 la
 #ComputeOvsDpdkInterface2=p3p1
 #BondInterfaceOvsOptions=bond_mode=balance-tcp lacp=active
 
+# To enable DVR with separate bond for Floating network on compute nodes, two interfaces should be used.  
+# To use the two interfaces, uncomment 'ComputeBond2Interface1' and 'ComputeBond2Interface2'.
```
heydog15: Shouldn't we also state to change the interfaces to match what is actually in the system since these are examples?

PR #191, opened 2018-11-01
Post deployment steps fixed (CES-10170)

beenum22:
- Added correct credentials required for Overcloud flavors and volume creation. 
- Host aggregation stack creation was not working. This step is redundant so removed it rather than fixing it.


heydog15:
Looks good.

PR #192, opened 2018-11-05
Request stacking

cdearborn:
Bumped up threads to avoid request stacking.


heydog15:
Looks good.  I think Gael has tested this successfully.

PR #195, opened 2018-11-14
Changes for OSP13z3

cdearborn:
This PR includes all the changes necessary to switch from OSP13z2 to OSP13z3.  It mostly consists of removing patches that are now available upstream.


cdearborn:
Merging this with permission from Kurt.

heydog15:
This looks good.  I was testing this with the Z3 puddle bits and all worked fine

PR #198, opened 2018-11-16
Fix RAID configuration on storage nodes

cdearborn:
Fix RAID configuration on storage nodes

This fixes an issue where if there user tells assign_role to do RAID
configuration, and also gives it the size of the volume to install on,
then we will still create the RAID volume according to our normal rules.

The prior behavior was to delete the RAID volume, convert the drives to
JBOD, and then abort RAID configuration.


jtwill98:
Less code, LGTM -- +1

PR #202, opened 2018-11-20
Added support to retrieve system properties from upstream python-dracclient

uday344:

1. Removed existing code from discover_nodes.
2. Added code to support retrieval of system properties from upstream python-dracclient.


src/pilot/config_idrac.py
```diff
@@ -68,10 +69,16 @@ def parse_arguments():
     return parser.parse_args()
 
 
-def get_pxe_nic_fqdd(fqdd, model_properties, drac_client):
```
cdearborn: I still don't understand why it is necessary to change this method signature.  Lines 76-81 below just repeat the code in the calling method.  We already have a drac_client created in the caller.  I can't see any reason why we wouldn't just pass that in here and use it.

src/pilot/assign_role.py
```diff
@@ -863,7 +858,6 @@ def place_node_in_available_state(ironic_client, node_uuid):
 def assign_role(ip_mac_service_tag, node_uuid, role_index, os_volume_size_gb,
                 ironic_client, drac_client):
     flavor = ROLES[role_index.role]
-
```
cdearborn: Can you add this blank line back in?

cdearborn: If pep8 was complaining about it, then the change is fine.

src/pilot/config_idrac.py
cdearborn: Oh, I see.  We're not planning on pushing those methods upstream.  We should change the code to use set_nic_settings() instead.

uday344:
@cdearborn 
I have resolved nic related dependencies as suggested above, but now I am getting following 

drac_client.wait_until_idrac_is_reset(False)
AttributeError: 'DRACClient' object has no attribute 'wait_until_idrac_is_reset'

For above I found a solution 

https://github.com/openstack/python-dracclient/commit/f823194b2e8892d8512df89b1d8e507373d59160

I have integrated above in the code but this method does not exist in 

https://pypi.org/project/python-dracclient/

So currently the code is not working due to above.
When are we planning to push the current code from github to pypi?

cdearborn:
@uday344, it probably is about time for us to cut a new release of python-dracclient.  This will cause a new version to be published to pypi.  I'll look in to doing that.  In the meantime, you could just grab the latest python-dracclient from github.

cdearborn:
I just cut a new release of python-dracclient.  2.0.0 is the one you want:
https://pypi.org/project/python-dracclient/2.0.0/

PR #205, opened 2018-11-21
Add support for compute nodes with only a boot disk

Rachit7194:
@cdearborn @rpioso @digambar15  Done suggested changes, please review it.


src/pilot/assign_role.py
```diff
@@ -321,10 +321,10 @@ def define_compute_logical_disks(drac_client, raid_controller_name):
     raid_10_logical_disk = define_single_raid_10_logical_disk(
         drac_client, raid_controller_name)
 
-    if raid_10_logical_disk is None:
-        return None
-
     logical_disks = list()
+    if not raid_10_logical_disk:
```
cdearborn: I don't think this is right.  define_single_raid_10_logical_disk can return either an empty dict or None.  None is meant to indicate that an error occurred, which should cause assign_role to exit.  The way this code is written, dict() and None will enter this "if" clause and return an empty list in both of those scenarios.  This will cause define_target_raid_config to return a dictionary in both cases, and that will cause this "if" statement to be entered (https://github.com/dsp-jetpack/JetPack/pull/205/files#diff-a1208342b5ece66e825ba657f52e8a16R783), and assign_role will continue along even in the error case.
I think this code should return a None if a None is returned from define_single_raid_10_logical_disk, and an empty list if an empty dict is returned, instead of returning an empty list in both cases.  I believe that would cause assign_role to abort in the event that an error occurs resulting in None being returned.

src/pilot/assign_role.py
```diff
@@ -308,10 +308,10 @@ def define_controller_logical_disks(drac_client, raid_controller_name):
...
```
cdearborn: See my comment below in define_compute_logical_disks.  The same holds true here.

src/pilot/assign_role.py
Rachit7194: As you suggested, I have modified the changes on current PR. Please review the same.

src/pilot/assign_role.py
Rachit7194: Done the changes. Please review it.

src/pilot/assign_role.py
```diff
@@ -308,11 +308,13 @@ def define_controller_logical_disks(drac_client, raid_controller_name):
     raid_10_logical_disk = define_single_raid_10_logical_disk(
         drac_client, raid_controller_name)
 
-    if raid_10_logical_disk is None:
-        return None
-
     logical_disks = list()
-    logical_disks.append(raid_10_logical_disk)
+    if raid_10_logical_disk is None:
+        return logical_disks
```
cdearborn: I'm pretty sure that on the line above, you need to:
return None
otherwise, assign_role won't exit with an error.  Can you make this change and test the error case to make sure this works correctly?  To test, you would need to hack define_single_raid_10_logical_disk() so that number_physical_disks=0 here:
https://github.com/dsp-jetpack/JetPack/blob/master/src/pilot/assign_role.py#L337
In this case, you should see a critical error logged here:
https://github.com/dsp-jetpack/JetPack/blob/master/src/pilot/assign_role.py#L373
and assign_role should exit with a 1 exit code here:
https://github.com/dsp-jetpack/JetPack/blob/master/src/pilot/assign_role.py#L1530

src/pilot/assign_role.py
```diff
@@ -321,11 +323,13 @@ def define_compute_logical_disks(drac_client, raid_controller_name):
     raid_10_logical_disk = define_single_raid_10_logical_disk(
         drac_client, raid_controller_name)
 
+    logical_disks = list()
     if raid_10_logical_disk is None:
         return None
```
cdearborn: The above line looks good.  Please test as noted above, but passing a role of "compute" instead.

src/pilot/assign_role.py
```diff
@@ -1161,6 +1166,23 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                    disks = [get_size_in_bytes(physical_disk_doc,
+                                               DCIM_PhysicalDiskView)
+                             for physical_disk_doc in physical_disk_docs]
+                    os_volume_size_gb = int(disks[0]) / units.Gi
+                except:
+                    raise RuntimeError(
+                        "There must be either a virtual disk that "
+                        "is not a RAID 0 to install the OS on, or "
+                        "os-volume-size-gb must be specified")
                 raise RuntimeError("There must be either a virtual disk that "
```
cdearborn: I don't see how this could possibly work because the above raise will be executed immediately following the try/except block, which means that an exception will always be thrown in the case where there is a jbod disk and no virtual disk.

src/pilot/assign_role.py
```diff
@@ -1161,6 +1166,23 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                        'DCIM_PhysicalDiskView',
+                        DCIM_PhysicalDiskView,
+                        True)
+                    disks = [get_size_in_bytes(physical_disk_doc,
+                                               DCIM_PhysicalDiskView)
+                             for physical_disk_doc in physical_disk_docs]
+                    os_volume_size_gb = int(disks[0]) / units.Gi
+                except:
+                    raise RuntimeError(
+                        "There must be either a virtual disk that "
```
cdearborn: Can we update this error message so that it's more accurate?  Something like:
"There must be a virtual disk that is not a RAID 0 or a single JBOD disk, or os-volume-size-gb must be specified."

src/pilot/assign_role.py
```diff
@@ -1161,6 +1166,23 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                        DCIM_PhysicalDiskView)
+                    physical_disk_docs = utils.find_xml(
+                        physical_disk_view_doc,
+                        'DCIM_PhysicalDiskView',
+                        DCIM_PhysicalDiskView,
+                        True)
+                    disks = [get_size_in_bytes(physical_disk_doc,
+                                               DCIM_PhysicalDiskView)
+                             for physical_disk_doc in physical_disk_docs]
+                    os_volume_size_gb = int(disks[0]) / units.Gi
```
cdearborn: Can we check to see if there is more than 1 disk, and raise an exception if so?

digambar15: I think it would be nice if we can only update the os_volume_size_gb for single disk. and add a note starting of block like "Below code block represents single disk scenario"

if len(disks) == 1:
     os_volume_size_gb = int(disks[0]) / units.Gi

src/pilot/assign_role.py
Rachit7194: Updated in latest commit. please review it.

src/pilot/assign_role.py
Rachit7194: Yes i have removed already.

src/pilot/assign_role.py
```diff
@@ -780,8 +784,9 @@ def configure_raid(ironic_client, node_uuid, role, os_volume_size_gb,
     target_raid_config = define_target_raid_config(
         role, drac_client)
 
-    if target_raid_config is None:
-        return False
+    if not target_raid_config or target_raid_config is None:
```
cdearborn: This "if" statement is returning success when target_raid_config is None, which is converting a critical failure into success.  When a critical failure occurs, assign_role should abort, and the way to make that happen is when target_raid_config is None, return False.  Only when the target_raid_config dict contains a "logical_disk" key that maps to an empty list, should True be returned here.

src/pilot/assign_role.py
```diff
@@ -1160,10 +1165,26 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
 
                     break
 
+            # Note: This code block represents single disk scenario.
             if raid_size_gb == 0:
-                raise RuntimeError("There must be either a virtual disk that "
-                                   "is not a RAID 0 to install the OS on, or "
-                                   "os-volume-size-gb must be specified")
+                try:
```
cdearborn: I think we should remove this try/except block.  If we just let any exception that occurs when communicating with the iDRAC go up the stack, then it will be caught in main() and handled correctly.

src/pilot/assign_role.py
```diff
@@ -1160,10 +1165,26 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                        DCIM_PhysicalDiskView)
+                    physical_disk_docs = utils.find_xml(
+                        physical_disk_view_doc,
+                        'DCIM_PhysicalDiskView',
+                        DCIM_PhysicalDiskView,
+                        True)
+                    disks = [get_size_in_bytes(physical_disk_doc,
+                                               DCIM_PhysicalDiskView)
+                             for physical_disk_doc in physical_disk_docs]
+                    if len(disks) == 1:
```
cdearborn: In the case where there is not 1 disk, then the exception below should be raised.

src/pilot/assign_role.py
```diff
@@ -1160,10 +1165,26 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
 
                     break
 
+            # Note: This code block represents single disk scenario.
             if raid_size_gb == 0:
-                raise RuntimeError("There must be either a virtual disk that "
-                                   "is not a RAID 0 to install the OS on, or "
-                                   "os-volume-size-gb must be specified")
+                try:
+                    physical_disk_view_doc = drac_client.enumerate(
```
cdearborn: I was talking with Rick today, and we came up with a case that needs to be handled which requires a change to this logic.  This is the case where the server has a RAID controller that does not support JBOD mode.  In that case, a RAID0 will be created that contains a single disk.  As a result, I think the logic needs to be:
- Check the virtual disks and see if there is a single RAID0.  If so then use that.  If there is more than one RAID0 then raise an exception
- If there are no RAID0's, then check the physical disks and see if there is a single disk.  If so, then use that.  If there is 0 or more than 1 physical disk, then raise an exception.

src/pilot/assign_role.py
cdearborn: Looks good

src/pilot/assign_role.py
digambar15: Chris - If there is single disk which is RAID0 and single physical disk without raid, anyway our code is taking the disk size irrespective of the disk type. If this is the case, then is it ok if we remove the RAID0 code for single disk ?
I didn't understand the RAID0 scenario for single disk which you are refering.
May be we can discuss this in today's meeting.

src/pilot/assign_role.py
```diff
@@ -781,7 +785,12 @@ def configure_raid(ironic_client, node_uuid, role, os_volume_size_gb,
         role, drac_client)
 
     if target_raid_config is None:
-        return False
+        return None
```
cdearborn: While the above line would work, I think it would be better to stick with a boolean and return False.

src/pilot/assign_role.py
```diff
@@ -781,7 +785,12 @@ def configure_raid(ironic_client, node_uuid, role, os_volume_size_gb,
...
+
+    if "logical_disks" in target_raid_config and not\
```
cdearborn: Looking at define_target_raid_config, it will either return None or a dictionary with a 'logical_disks' key.  As a result, I think you can simplify the above line to:
if not target_raid_config['logical_disks']:

src/pilot/assign_role.py
Rachit7194: Done.

src/pilot/assign_role.py
Rachit7194: To be on safe side i checked `logical_disks` key present or not in target_raid_config. But now as we split the condition to check None in target_raid_config, i think `if not target_raid_config['logical_disks']:` can works. so removed as you suggested.

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,46 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
 
                     break
 
+            # Note: This code block represents single disk scenario.
+            raid_disk_ids = []
```
cdearborn: Would you mind moving the above line to just after the "if virtual_disk_docs:" line?

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,46 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                            DCIM_VirtualDiskView).text
+
+                        if raid_type == RAID0:
+                            raid_size = get_size_in_bytes(virtual_disk_doc,
+                                                          DCIM_VirtualDiskView)
+                            raid_size_gb = int(raid_size) / units.Gi
+                            raid_disk_ids.append(raid_size_gb)
+                    if not len(raid_disk_ids) == 1:
+                        raise RuntimeError(
+                            "There must be either a virtual disk that "
```
cdearborn: Can you change this error message to: "There must be a non-RAID0 virtual disk, a single disk RAID0, or a single JBOD disk to install the OS on, or os-volume-size-gb must be specified."

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,46 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                        physical_disk_view_doc,
+                        'DCIM_PhysicalDiskView',
+                        DCIM_PhysicalDiskView,
+                        True)
+                    disks = [get_size_in_bytes(physical_disk_doc,
+                                               DCIM_PhysicalDiskView)
+                             for physical_disk_doc in physical_disk_docs]
+                    if not len(disks) == 1:
+                        raise RuntimeError(
+                            "There must be either a virtual disk that "
```
cdearborn: Can you change this error message to: "There must be a non-RAID0 virtual disk, a single disk RAID0, or a single JBOD disk to install the OS on, or os-volume-size-gb must be specified."

cdearborn:
I have 3 very minor comments, and I think this will be good.  Thanks!

src/pilot/assign_role.py
Rachit7194: Shifted `raid_disk_ids` list on when `virtual_disk_docs` exists.

src/pilot/assign_role.py
Rachit7194: Changed.

src/pilot/assign_role.py
Rachit7194: Changed.

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,49 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                             for physical_disk_doc in physical_disk_docs]
+                    if not len(disks) == 1:
+
+                        raise RuntimeError(
+                            "There must be a non-RAID0 virtual disk,"
+                            "a single disk RAID0, or a single JBOD disk"
+                            "to install the OS on,"
+                            "or os-volume-size-gb must be specified.")
 
+                    os_volume_size_gb = int(disks[0]) / units.Gi
```
rpioso: Does the check below to see if there are any physical disks that do not back the RAID that are the same size as the RAID need to be accommodated by this new functionality? I do not see `raid_physical_disk_ids` being set.

rpioso:
During my code review, I reconciled the changes in this pull request against the guidance @cdearborn and I provided in an e-mail. That e-mail was sent by me to @digambar15, copying @cdearborn, on Thursday, 11/15/2018, 4:53 EST, with subject "RE: Single JBOD disk support".

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,49 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                if virtual_disk_docs:
+                    raid_disk_ids = []
+                    for virtual_disk_doc in virtual_disk_docs:
+                        fqdd = get_fqdd(virtual_disk_doc, DCIM_VirtualDiskView)
+                        raid_type = utils.find_xml(
+                            virtual_disk_doc,
+                            'RAIDTypes',
+                            DCIM_VirtualDiskView).text
+
+                        if raid_type == RAID0:
```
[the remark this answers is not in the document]
digambar15: Lets discuss this in meeting.

src/pilot/assign_role.py
Rachit7194: i will add physical disk  which is equal to `RAID0` type to `raid_physical_disk_ids` list

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,49 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                                                          DCIM_VirtualDiskView)
+                            raid_size_gb = int(raid_size) / units.Gi
+                            raid_disk_ids.append(raid_size_gb)
+                    if not len(raid_disk_ids) == 1:
+                        raise RuntimeError(
+                            "There must be a non-RAID0 virtual disk,"
+                            "a single disk RAID0, or a single JBOD disk"
+                            "to install the OS on,"
+                            "or os-volume-size-gb must be specified.")
+                    os_volume_size_gb = raid_disk_ids[0]
```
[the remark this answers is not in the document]
Rachit7194: i will replace it to 'raid_size_gb` from `os_volume_size_gb `

src/pilot/assign_role.py
```diff
@@ -1389,7 +1435,6 @@ def change_physical_disk_state(drac_client, mode,
                     physical_disk.controller]
 
                 physical_disk_ids.append(physical_disk.id)
-
```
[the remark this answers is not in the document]
Rachit7194: Actually i did not find `-` on line 1438 of assign_role.py script.

src/pilot/assign_role.py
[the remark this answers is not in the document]
Rachit7194: ok

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,49 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                                                 DCIM_PhysicalDiskView)
+                    physical_disk_docs = utils.find_xml(
+                        physical_disk_view_doc,
+                        'DCIM_PhysicalDiskView',
+                        DCIM_PhysicalDiskView,
+                        True)
+                    disks = [get_size_in_bytes(physical_disk_doc,
+                                               DCIM_PhysicalDiskView)
+                             for physical_disk_doc in physical_disk_docs]
+                    if not len(disks) == 1:
```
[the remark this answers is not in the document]
Rachit7194: i will change it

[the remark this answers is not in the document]
Rachit7194: i will change it

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,49 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                        raid_type = utils.find_xml(
+                            virtual_disk_doc,
+                            'RAIDTypes',
+                            DCIM_VirtualDiskView).text
+
+                        if raid_type == RAID0:
+                            raid_size = get_size_in_bytes(virtual_disk_doc,
+                                                          DCIM_VirtualDiskView)
+                            raid_size_gb = int(raid_size) / units.Gi
+                            raid_disk_ids.append(raid_size_gb)
```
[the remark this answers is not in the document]
Rachit7194: ok

src/pilot/assign_role.py
```diff
@@ -308,11 +308,13 @@ def define_controller_logical_disks(drac_client, raid_controller_name):
...
         return None
-
-    logical_disks = list()
-    logical_disks.append(raid_10_logical_disk)
+    elif isinstance(raid_10_logical_disk, dict) and not raid_10_logical_disk:
+        return logical_disks
+    else:
+        logical_disks.append(raid_10_logical_disk)
 
     return logical_disks
```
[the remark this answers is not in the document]
Rachit7194: yes it seems more simplified..thanks

src/pilot/assign_role.py
```diff
@@ -321,11 +323,13 @@ def define_compute_logical_disks(drac_client, raid_controller_name):
...
```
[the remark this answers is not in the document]
Rachit7194: sure

src/pilot/assign_role.py
```diff
@@ -1160,11 +1168,49 @@ def select_os_volume(os_volume_size_gb, ironic_client, drac_client, node_uuid):
...
+                            virtual_disk_doc,
+                            'RAIDTypes',
+                            DCIM_VirtualDiskView).text
+
+                        if raid_type == RAID0:
+                            raid_size = get_size_in_bytes(virtual_disk_doc,
+                                                          DCIM_VirtualDiskView)
+                            raid_size_gb = int(raid_size) / units.Gi
+                            raid_disk_ids.append(raid_size_gb)
+                    if not len(raid_disk_ids) == 1:
```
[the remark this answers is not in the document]
Rachit7194: i will change it.

src/pilot/assign_role.py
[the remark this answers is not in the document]
Rachit7194: i have not added this change in latest commit because need to discuss this in meeting.

cdearborn:
The file permissions on assign_role.py were changed from 755 to 644, which means that executable permissions have been removed.  Please add executable permissions back in.

src/pilot/assign_role.py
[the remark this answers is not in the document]
cdearborn: I think what rpioso meant is that the above blank line was deleted by this PR, but that change was not necessary.  If this resolves a pep8 issue, then please leave as-is.  If it does not, then please add the blank line back in.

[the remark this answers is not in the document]
Rachit7194: It was giving pep8 error as `to many blank lines` so i removed it

src/pilot/assign_role.py
```diff
@@ -296,11 +296,15 @@ def define_controller_logical_disks(drac_client, raid_controller_name):
...
 
+    # None indicates an error occurred.
     if raid_10_logical_disk is None:
         return None
 
     logical_disks = list()
-    logical_disks.append(raid_10_logical_disk)
+
+    # Add the disk to the list only if it is not empty.
+    if logical_disks:
```
dpaterson: This doesn't make sense as logical_disks is always initialized to empty list this code never executes, I think you are supposed to check for  if raid_10_logical_disks: ?

src/pilot/assign_role.py
```diff
@@ -309,11 +313,15 @@ def define_compute_logical_disks(drac_client, raid_controller_name):
...
+        logical_disks.append(raid_10_logical_disk)
```
dpaterson: Same as above, logical_disks is always empty as it is initialized on line #320 above

src/pilot/assign_role.py
Rachit7194: Oops..!!..Apologies and corrected the same.

src/pilot/assign_role.py
Rachit7194: Changed.

Rachit7194:
I have tested the above changes including the latest commits on both the nodes compute and controller respectively.

PR #207, opened 2018-11-22
locking rpms & containers

gaelrehault:


heydog15:
Looks like what I created for the director and dashboard

PR #210, opened 2018-11-28
Modified reset_idrac method to point to Upstream reset_idrac method.

uday344:
@cdearborn @digambar15 @rpioso 
Please review and let me know the changes if required


digambar15:
We had to take this approach because reset_idrac was existed in the downstream dracclient also and whenever we invoke, downstream method was called. so we have delete downstream reset_idrac method to talk to upstream dracclient method.

uday344:
Removed wait_until_idrac_is_reset from client.py and reset_idrac from idrac_card.py

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
```diff
@@ -77,7 +77,7 @@ def __init__(self,
                                          path,
                                          protocol)
         self._job_mgmt = job.JobManagement(self.client)
-        self._idrac_cfg = idrac_card.iDRACCardConfiguration(self.client)
+        self._idrac_con = idrac_card.iDRACCardConfiguration(self.client)
```
rpioso: Now that `config_idrac.py` uses upstream's `reset_idrac()`, does `discover_nodes.resources.idrac_card` add any value? If not, this member variable and that module could be removed now **and** resolve a puzzling situation, rather than perpetuate it.

Thoughts?

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
```diff
@@ -93,7 +93,7 @@ def list_idrac_settings(self):
         :raises: DRACOperationFailed on error reported back by the DRAC
                  interface
         """
-        return self._idrac_cfg.list_idrac_settings()
+        return self._idrac_con.list_idrac_settings()
```
rpioso: Please see my comment above about the `_idrac_con` data member. If that's removed, this method must be removed, too.

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
```diff
@@ -116,124 +116,7 @@ def set_idrac_settings(self, settings, idrac_fqdd='iDRAC.Embedded.1'):
...
-            retries -= 1
-            sleep(10)
-
-        if retries == 0:
-            raise exceptions.DRACOperationFailed(drac_messages="Timed out "
-                                                 "waiting for the " +
-                                                 self.client.host +
-                                                 " iDRAC to "
-                                                 "become ready")
+        return self._idrac_con.set_idrac_settings(idrac_fqdd, settings)
```
rpioso: Please see my comment above about the `_idrac_con` data member. If that's removed, this method, `set_idrac_settings()`, must be removed, too.

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
uday344: Richard, Earlier I tried removing downstream drac client from config_idrac.py but faced multiple dependency issues, So we decided to go task by task which may need to add some temporary changes to local downstream drac client but eventually, all the repeated and unused code will be removed.
This way we avoid complications in code and it will be easy to review the code.

I currently facing some issues regarding dependencies of set_idrac_settings method.
Getting 
"InvalidParameterValue: Unknown iDRAC Card attributes found: set(['A', 'C', 'b', 'E', 'D', 'i', 'm', '.', '1', 'R', 'e', 'd'])"

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
uday344: Done, Please review

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
uday344: Done, Please review

src/pilot/discover_nodes/discover_nodes/dracclient/client.py
uday344: Modified this PR with requested changes, Please review

PR #213, opened 2018-12-04
updated sample switch config

gaelrehault:


data/sample_switch_configs/5-NICS_S4048-config
```diff
@@ -0,0 +1,157 @@
+# This file describes the network interfaces available on your system
+# and how to activate them. For more information, see interfaces(5).
```
heydog15: What does it mean "see interfaces(5)"?

data/sample_switch_configs/5-NIC_s5248-1-config
```diff
@@ -0,0 +1,317 @@
+# This file describes the network interfaces available on your system
+# and how to activate them. For more information, see interfaces(5).
```
heydog15: Same comment as the 4048

data/sample_switch_configs/5-NIC_s5248-2-config
```diff
@@ -0,0 +1,350 @@
+# This file describes the network interfaces available on your system
+# and how to activate them. For more information, see interfaces(5).
```
heydog15: Same comment

PR #216, opened 2018-12-05
Use upstreamed is_jbod_capable CES-10113

jainprasuk1996:
- Removed is_jbod_capable method from assign_role.py.
- Used is_jbod_capable method from upstream in assign_role.py.
@cdearborn @digambar15

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -31,6 +31,7 @@ import yaml
 from dracclient import utils
 from dracclient import client
 from dracclient.constants import POWER_OFF
+from dracclient.resources import raid
 from dracclient.exceptions import DRACOperationFailed, \
     DRACUnexpectedReturnValue, WSManInvalidResponse, WSManRequestFailure
 from oslo_utils import units
@@ -428,7 +429,7 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
     #
     # A successful call returns a list, which may be empty; otherwise,
     # None is returned.
-    jbod_capable = is_jbod_capable(drac_client, raid_controller_name)
+    jbod_capable = raid.is_jbod_capable(drac_client, raid_controller_name)
     jbod_logical_disks = define_jbod_logical_disks(
         drac_client, remaining_physical_disks, raid_controller_name,
         jbod_capable)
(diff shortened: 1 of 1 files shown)


digambar15:
Looks Good

src/pilot/assign_role.py
```diff
@@ -428,7 +429,7 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
     #
     # A successful call returns a list, which may be empty; otherwise,
     # None is returned.
-    jbod_capable = is_jbod_capable(drac_client, raid_controller_name)
+    jbod_capable = raid.is_jbod_capable(drac_client, raid_controller_name)
```
cdearborn: We should be going through the client object here instead of calling the method in the resource class.  Like this:
drac_client.is_jbod_capable(raid_controller_name)

src/pilot/assign_role.py
```diff
@@ -619,7 +620,7 @@ def define_jbod_or_raid_0_logical_disk(drac_client,
                                        is_root_volume=False,
                                        jbod_capable=None):
     if jbod_capable is None:
-        jbod_capable = is_jbod_capable(drac_client, raid_controller_name)
+        jbod_capable = raid.is_jbod_capable(drac_client, raid_controller_name)
```
cdearborn: Same comment as above.

src/pilot/assign_role.py
```diff
@@ -31,6 +31,7 @@
 from dracclient import utils
 from dracclient import client
 from dracclient.constants import POWER_OFF
+from dracclient.resources import raid
```
cdearborn: Once the change has been made below to go through the client object, I don't think this import will be needed any more.

src/pilot/assign_role.py
jainprasuk1996: copy

src/pilot/assign_role.py
jainprasuk1996: copy

src/pilot/assign_role.py
jainprasuk1996: copy

src/pilot/assign_role.py
jainprasuk1996: done

src/pilot/assign_role.py
jainprasuk1996: done

src/pilot/assign_role.py
jainprasuk1996: done

rpioso:
The changes look good.

I have a question before I approve this pull request. Does the version of python-dracclient that is included in Red Hat OpenStack Platform (OSP) 13 include [Change 578564 - Added is_jbod_capable method to raid resource](https://review.openstack.org/#/c/578564/)?

cdearborn:
@rpioso , yes, is_jbod_capable is present in the version of python-dracclient that is shipping with OSP13z3.

PR #219, opened 2018-12-12
Removed unnecessary changes from upstreamed system properties

uday344:
This PR is an extension of https://github.com/dsp-jetpack/JetPack/pull/202

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -29,7 +29,6 @@ import sys
 import yaml
 
 from dracclient import utils
-from dracclient import client
 from dracclient.constants import POWER_OFF
 from dracclient.exceptions import DRACOperationFailed, \
     DRACUnexpectedReturnValue, WSManInvalidResponse, WSManRequestFailure
@@ -42,6 +41,12 @@ from logging_helper import LoggingHelper
 import requests.packages
 from ironicclient.common.apiclient.exceptions import InternalServerError
 
+discover_nodes_path = os.path.join(os.path.expanduser('~'),
+                                   'pilot/discover_nodes')
+
+sys.path.append(discover_nodes_path)
+
+from discover_nodes.dracclient.client import DRACClient  # noqa
 requests.packages.urllib3.disable_warnings()
 
 # Perform basic configuration of the logging system, which configures the root
@@ -224,7 +229,7 @@ def get_drac_client(node_definition_filename, node):
     drac_ip, drac_user, drac_password = \
         CredentialHelper.get_drac_creds_from_node(node,
                                                   node_definition_filename)
-    drac_client = client.DRACClient(drac_ip, drac_user, drac_password)
+    drac_client = DRACClient(drac_ip, drac_user, drac_password)
     # TODO: Validate the IP address is an iDRAC.
     #
     #       This could detect an error by an off-roading user who provided an
--- a/src/pilot/discover_nodes/discover_nodes.py
+++ b/src/pilot/discover_nodes/discover_nodes.py
@@ -19,7 +19,6 @@ from __future__ import print_function
 
 import argparse
 from collections import namedtuple
-from dracclient import client
 from exceptions import ValueError
 import json
 import logging
@@ -30,6 +29,8 @@ import dracclient.exceptions
 import dracclient.utils
 import netaddr
 import requests.packages
+
+import discover_nodes.dracclient.client as discover_nodes_dracclient
 
 # Suppress InsecureRequestWarning: Unverified HTTPS request is being
 # made. See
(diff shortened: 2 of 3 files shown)


src/pilot/assign_role.py
```diff
@@ -42,6 +41,12 @@
 import requests.packages
 from ironicclient.common.apiclient.exceptions import InternalServerError
 
+discover_nodes_path = os.path.join(os.path.expanduser('~'),
+                                   'pilot/discover_nodes')
+
+sys.path.append(discover_nodes_path)
+
+from discover_nodes.dracclient.client import DRACClient  # noqa
```
digambar15: Let's stick to - 
from discover_nodes.dracclient import client  
to be consistent with other imports.

src/pilot/assign_role.py
```diff
@@ -224,7 +229,7 @@ def get_drac_client(node_definition_filename, node):
     drac_ip, drac_user, drac_password = \
         CredentialHelper.get_drac_creds_from_node(node,
                                                   node_definition_filename)
-    drac_client = client.DRACClient(drac_ip, drac_user, drac_password)
+    drac_client = DRACClient(drac_ip, drac_user, drac_password)
```
digambar15: if we add above import change, then we dont require change here

src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -31,6 +30,8 @@
 import netaddr
 import requests.packages
 
+import discover_nodes.dracclient.client as discover_nodes_dracclient
```
digambar15: Ditto

src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -366,9 +367,9 @@ def scan_one(scan_info):
     # protocol. See the DMTF's "Web Services for Management
     # (WS-Management) Specification"
     # (http://www.dmtf.org/sites/default/files/standards/documents/DSP0226_1.2.0.pdf).
-    drac_client = client.DRACClient(scan_info.ip_address,
-                                    scan_info.user_name,
-                                    scan_info.password)
+    client = discover_nodes_dracclient.DRACClient(scan_info.ip_address,
```
digambar15: Same as above

src/pilot/assign_role.py
cdearborn: Can you explain why we are making the change so that assign_role uses the discover_nodes DRACClient instead of the DRACClient from python-dracclient?  Ideally, we want to eliminate the discover_nodes DRACClient and switch everything to the python-dracclient DRACClient.

src/pilot/discover_nodes/discover_nodes.py
cdearborn: Same here.  I don't understand why we are switching to DRACClient in discover_nodes.

cdearborn:
The changes in config_idrac.py looks good (thanks for remembering to do that).  I'm not sure about the changes to assign_role and discover_nodes though.

src/pilot/assign_role.py
uday344: My intention behind this PR was to remove extra added dracclients (which was added in https://github.com/dsp-jetpack/JetPack/pull/202) and to make fewer changes in code.

In assign_role.py 

I have removed upstream dracclient and reverted back downstream dracclient (Same as Master).
https://github.com/dsp-jetpack/JetPack/blob/8cff5fd7c2ed90876d4cf851862a30bf756480d7/src/pilot/assign_role.py#L49

Above changes eventually point to 
https://github.com/dsp-jetpack/JetPack/blob/8cff5fd7c2ed90876d4cf851862a30bf756480d7/src/pilot/assign_role.py#L904

Earlier this change is done via newly imported upstream dracclient, but now it is working from downstream via inheritance.

Again this is up to us if we want to incorporate these changes.

src/pilot/discover_nodes/discover_nodes.py
uday344: As stated in the above comment this change is also done for the same purpose.
Removed upstream dracclient and reverted back downstream dracclient (same as master).
This changes eventually point to https://github.com/dsp-jetpack/JetPack/blob/8cff5fd7c2ed90876d4cf851862a30bf756480d7/src/pilot/discover_nodes/discover_nodes.py#L390 and https://github.com/dsp-jetpack/JetPack/blob/8cff5fd7c2ed90876d4cf851862a30bf756480d7/src/pilot/discover_nodes/discover_nodes.py#L391

Earlier this change is done via newly imported upstream dracclient, but now it is working from downstream via inheritance.

uday344:
My intention behind this PR was to remove extra added dracclients (which was added in #202) and to make fewer changes in code as compared to Master's Code.

Please note that this PR is optional and it does not have any dependencies, this can be considered as code optimizations and improvements.

src/pilot/assign_role.py
dpaterson: question: what is #noqa comment for?

PR #222, opened 2018-12-13
 CES-10114- Use upstreamed conversion of drives from RAID to JBOD and back

Rachit7194:
@rpioso @cdearborn  @digambar15  Guys, I have done the respective changes please review it


src/pilot/assign_role.py
```diff
@@ -262,32 +262,20 @@ def define_target_raid_config(role, drac_client):
...
-        c.id for c in disk_ctrls if is_boss_controller(c.id, drac_client)
+        c.id for c in disk_ctrls if drac_client.is_boss_controller(c.id)
     ]
 
     if boss_controller_ids:
         raid_controller_ids.append(boss_controller_ids[0])
     number_raid_controllers = len(raid_controller_ids)
 
-    if number_raid_controllers == 1:
+    if number_raid_controllers >= 1:
```
digambar15: @cdearborn  - Need discussion on this condition. We decided to go with this approach because we are getting more than 1 controllers always for any node.

src/pilot/assign_role.py
```diff
@@ -1288,14 +1276,21 @@ def ensure_node_is_powered_off(drac_client):
...
         controllers_to_physical_disk_ids=None):
-    reboot_required, job_ids = change_physical_disk_state(
-        drac_client, mode, controllers_to_physical_disk_ids)
+
+    physical_disk_status = drac_client.change_physical_disk_state(
+        mode, controllers_to_physical_disk_ids)
+
+    job_ids = []
+    if physical_disk_status['commit_required_ids']:
+        for controller in physical_disk_status['commit_required_ids']:
```
cdearborn: Would you mind changing "controller" to "controller_id" above?  There is a controller object, but the variable above actually contains just an id and not the object.

Rachit7194: sure i will do the change

src/pilot/assign_role.py
```diff
@@ -1288,14 +1276,21 @@ def ensure_node_is_powered_off(drac_client):
 def change_physical_disk_state_wait(
         node_uuid, ironic_client, drac_client, mode,
         controllers_to_physical_disk_ids=None):
-    reboot_required, job_ids = change_physical_disk_state(
-        drac_client, mode, controllers_to_physical_disk_ids)
+
+    physical_disk_status = drac_client.change_physical_disk_state(
```
cdearborn: There is a PhysicalDisk.status attribute, so the above variable name is a bit confusing since at first glance, I thought it had something to do with that attribute.  Could we change the variable name to something else, like "change_state_result"?

Rachit7194:
I have tested this change with full deployment where undercloud creates again and upstream dracclient got successfully updated without deployment failure.
NOTE: Tested this deployment by skipping one node as it was having some configuration problem(not getting power on).

PR #225, opened 2018-12-18
Use upstreamed nic configuration methods and remove downstream counterparts

jainprasuk1996:

Note:
Closed the previous PR for the same story due to some problem in it. This is a new PR which contains all the changes.


digambar15:
Thanks for approaving this PR.
Sure Chris. We will come up with another PR which will be targeted for corresponding changes.

jainprasuk1996:
Thanks.

PR #226, opened 2018-12-19
Remove deprecated commit_required parameter from JetPack

Rachit7194:
I have removed `commit_required `parameter with `is_commit_required` and `is_reboot_required `as per the given condition respectively. Guys @rpioso @digambar15  @cdearborn please review it.

### DIFF
--- a/src/pilot/config_idrac.py
+++ b/src/pilot/config_idrac.py
@@ -215,8 +215,6 @@ def config_boot_mode(drac_client, ip_service_tag, node, boot_mode):
         job_id = drac_client.commit_pending_bios_changes(reboot=False,
                                                          start_time=None)
 
-    # Note that "commit_required" is actually "reboot_required" under the
-    # covers
     return response['is_reboot_required'], job_id
 
 


src/pilot/config_idrac.py
```diff
@@ -142,9 +142,9 @@ def configure_uefi_nics_boot_settings(drac_client, pxe_nic_id):
 
             provisioning_mac = nic.mac_address.lower()
             response = drac_client.set_bios_settings(settings)
-            reboot_required = response['commit_required']
+            reboot_required = response['is_reboot_required']
 
-            if response['commit_required']:
+            if response['is_commit_required']:
```
cdearborn: Would you mind changing the above line to:
if reboot_required:

cdearborn: I'm sorry, please ignore the above comment.  Line 147 is correct as it is.

cdearborn:
Looks good!  Please post a comment when testing of the assign_role change is complete.

Rachit7194:
I have tested the assign_role successfully and we are good to go.

src/pilot/config_idrac.py
```diff
@@ -211,13 +211,13 @@ def config_boot_mode(drac_client, ip_service_tag, node, boot_mode):
     response = drac_client.set_bios_settings(settings)
 
     job_id = None
-    if response['commit_required']:
+    if response['is_commit_required']:
         job_id = drac_client.commit_pending_bios_changes(reboot=False,
                                                          start_time=None)
 
     # Note that "commit_required" is actually "reboot_required" under the
```
dpaterson: This comment is confusing now, should it be removed?

cdearborn: yes.  since commit_required has been removed, we should remove this comment too.

PR #228, opened 2019-01-03
Enforce IB introspection for CSP profile

cdearborn:
Evidently NUMA and hugepages require IB introspection, so enforce IB introspection for the CSP profile.


heydog15:
Looks good.

gaelrehault:
Looks good, tho we probably should check for this condition in code as well to allow for custom profiles

PR #230, opened 2019-01-09
Added ability to skip_baremetal_configuration (NIC and BIOS)

jainprasuk1996:
Note: A few changes are just for the sake of PEP8
@digambar15 @cdearborn @rpioso

### DIFF
--- a/src/deploy/osp_deployer/settings/config.py
+++ b/src/deploy/osp_deployer/settings/config.py
@@ -572,6 +572,18 @@ class Settings():
                 except AttributeError:
                     node.skip_raid_config = False
                     pass
+                try:
+                    if node.skip_bios_config == "true":
+                        node.skip_bios_config = True
+                except AttributeError:
+                    node.skip_bios_config = False
+                    pass
+                try:
+                    if node.skip_nic_config == "true":
+                        node.skip_nic_config = True
+                except AttributeError:
+                    node.skip_nic_config = False
+                    pass
 
         Settings.settings = self
 
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -1478,6 +1478,8 @@ def main():
 
                 if not succeeded:
                     sys.exit(1)
+            else:
+                LOG.info("Skipping RAID configuration")
 
             if not args.skip_bios_config:
                 succeeded = configure_bios(
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -123,6 +123,10 @@ def parse_arguments():
                         '--skip-raid-config',
                         action='store_true',
                         help="skip configuring RAID")
+    parser.add_argument('-b',
+                        '--skip-bios-config',
+                        action='store_true',
+                        help="skip configuring BIOS")
     parser.add_argument('-o',
                         '--os-volume-size-gb',
                         help="the size of the volume to install the OS on "
@@ -1475,15 +1479,17 @@ def main():
                 if not succeeded:
                     sys.exit(1)
 
-            succeeded = configure_bios(
-                node,
-                ironic_client,
-                bios_settings,
-                drac_client)
+            if not args.skip_bios_config:
+                succeeded = configure_bios(
+                    node,
+                    ironic_client,
+                    bios_settings,
+                    drac_client)
 
-            if not succeeded:
-                sys.exit(1)
-
+                if not succeeded:
+                    sys.exit(1)
+            else:
+                LOG.info("Skipping BIOS configuration")
         assign_role(
             args.ip_mac_service_tag,
             node.uuid,
--- a/src/pilot/config_idrac.py
+++ b/src/pilot/config_idrac.py
@@ -133,7 +133,7 @@ def configure_nics_boot_settings(
                                                  pxe_nic_id)
 
 
-def skipped_configure_nics_boot_settings(drac_client, pxe_nic_id):
+def get_nic_mac_address(drac_client, pxe_nic_id):
     for nic in drac_client.list_nics(sort=True):
         # Compare the NIC IDs case insensitively. Assume ASCII strings.
         if nic.id.lower() == pxe_nic_id.lower():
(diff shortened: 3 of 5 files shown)


src/pilot/config_idrac.py
```diff
@@ -415,33 +427,48 @@ def config_idrac(instack_lock,
...
+                                         ip_service_tag,
+                                         pxe_nic_fqdd,
+                                         node,
+                                         target_boot_mode)
+
+        reboot_required = reboot_required or reboot_required_nic
+        if nic_job_ids:
+            job_ids.extend(nic_job_ids)
+
+        # Do initial idrac configuration
```
cdearborn: This section of code that deals with idrac configuration needs to be executed even if the user chooses to skip NIC configuration.  This is because the user may choose to change the password on the iDRAC, and that will only happen when config_idrac_settings is called.  I think you can just un-indent 455-471 to achieve this.

src/pilot/config_idrac.py
```diff
@@ -129,6 +133,13 @@ def configure_nics_boot_settings(
                                                  pxe_nic_id)
 
 
+def skipped_configure_nics_boot_settings(drac_client, pxe_nic_id):
```
cdearborn: Can you rename this method to something like "get_nic_mac_address"?

src/pilot/config_idrac.py
```diff
@@ -415,33 +427,48 @@ def config_idrac(instack_lock,
...
-    # If we need to reboot, then add a job for it
-    if reboot_required:
-        LOG.info("Rebooting {} to apply configuration".format(ip_service_tag))
-
-        job_id = drac_client.create_reboot_job()
-        job_ids.append(job_id)
+    boot_modes_same = BootModeHelper.get_boot_mode(
+        drac_client) == target_boot_mode
+
+    if skip_nic_config and boot_modes_same:
```
cdearborn: This story was originally created because an expert user configured the NICs manually and then wanted us to not change that configuration.  This was before UEFI support was added.  To meet the spirit of that story, I think we should NOT change the boot mode, but should leave it as configured when the user chooses to skip NIC configuration.  This is because the user could have configured the NICs as they wanted (how the PXE NIC is configured would be dependent on if the user pick UEFI or BIOS for the boot mode), but if we change the boot mode, that configuration will be lost.  As a result, I think we should change the code where target_boot_mode is initially set.  If skip_nic_configuration is specified, then we should just retrieve the boot mode from the server and set target_boot_mode to that.  This would make the boot_modes_same check unnecessary at this point in the code.

src/pilot/config_idracs.py
```diff
@@ -104,15 +105,19 @@ def main():
 
                     if "password" in node_config.keys():
                         password = node_config["password"]
-
+                    if "skip_nic_config" in json_config[
+                        node["pm_addr"]].keys():
+                        skip_nic_config = json_config[node["pm_addr"]
```
cdearborn: I think that the above line should be:
skip_nic_config = node_config["skip_nic_config"]
This is because the code above on lines 95-99 is pulling out the node dictionary based on either the ip address (pm_addr) or service tag, while the line above assumes that the node is always indexed by ip address.

src/pilot/assign_role.py
```diff
@@ -1475,15 +1479,17 @@ def main():
...
+                succeeded = configure_bios(
+                    node,
+                    ironic_client,
+                    bios_settings,
+                    drac_client)
 
+                if not succeeded:
+                    sys.exit(1)
+            else:
+                LOG.info("Skipping BIOS configuration")
```
cdearborn: I like that you have a logging statement here when skipping BIOS configuration.  Would you mind adding a similar logging statement above when RAID configuration is skipped?

src/pilot/config_idrac.py
jainprasuk1996: done

src/pilot/config_idrac.py
jainprasuk1996: done

src/pilot/config_idrac.py
jainprasuk1996: done

src/pilot/config_idracs.py
jainprasuk1996: done

src/pilot/assign_role.py
jainprasuk1996: done

src/pilot/config_idrac.py
```diff
@@ -415,33 +427,48 @@ def config_idrac(instack_lock,
...
+        provisioning_mac = skipped_configure_nics_boot_settings(
```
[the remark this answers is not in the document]
jainprasuk1996: Done

digambar15:
Looks good

src/pilot/config_idrac.py
```diff
@@ -129,6 +133,13 @@ def configure_nics_boot_settings(
                                                  pxe_nic_id)
 
 
+def get_nic_mac_address(drac_client, pxe_nic_id):
+    for nic in drac_client.list_nics(sort=True):
+        # Compare the NIC IDs case insensitively. Assume ASCII strings.
+        if nic.id.lower() == pxe_nic_id.lower():
+            return nic.mac_address.lower()
```
cdearborn: It doesn't look like the above could work because the method that constructs a NIC in python-dracclient has a mac attribute, not a mac_address attribute:
https://github.com/openstack/python-dracclient/blob/master/dracclient/resources/inventory.py#L181

jainprasuk1996:
The mac_address is an attribute for the downstream client. Currently, this code is referring to the downstream part itself. We ran a couple of E2E deployments after making these changes and it worked fine.

Is it possible that we merge this PR before I raise next PR for NIC configuration management story?

src/deploy/osp_deployer/settings/config.py
```diff
@@ -572,6 +572,18 @@ def __init__(self, settings_file):
...
+                        node.skip_bios_config = True
+                except AttributeError:
+                    node.skip_bios_config = False
+                    pass
+                try:
+                    if node.skip_nic_config == "true":
+                        node.skip_nic_config = True
+                except AttributeError:
+                    node.skip_nic_config = False
+                    pass
```
dpaterson: Might be nice to introduce the standard python library distutils to our stuff. Then we could simplify above into something like: node.skip_bios_config = bool(util.strtobool(node.skip_bios_config))
https://docs.python.org/2/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool

Would allow us to skip all the exception handling etc I believe.

PR #233, opened 2019-01-15
Nic configuration management

jainprasuk1996:
Removed JetPack nic.py file and pointed the code to upstreamed nic.py.
Added methods in config_idrac file which were not part of upstreamed nic.
@rpioso @cdearborn @digambar15


src/pilot/config_idrac.py
```diff
@@ -204,6 +207,199 @@ def configure_bios_nics_boot_settings(drac_client, ip_service_tag, pxe_nic_id):
...
+              setting to be applied
+    :raises: WSManRequestFailure on request failures
+    :raises: WSManInvalidResponse when receiving invalid response
+    :raises: DRACOperationFailed on error reported back by the iDRAC
+             interface
+    """
+    return set_nic_legacy_boot_protocol(nic_id, 'PXE', drac_client)
+
+
+def _list_nic_settings(selection_expression):
```
cdearborn: We should be using the upstreamed list_nic_settings() method and delete this one.

src/pilot/config_idrac.py
```diff
@@ -204,6 +207,199 @@ def configure_bios_nics_boot_settings(drac_client, ip_service_tag, pxe_nic_id):
...
+                'nic': nic_id})
+
+    # Do the settings include the attribute?
+    if attribute_name not in settings:
+        return None
+
+    return settings[attribute_name]
+
+
+def set_nic_setting(nic_id, attribute_name, value, drac_client):
```
cdearborn: We should be using the upstreamed set_nic_settings() method and delete this one.

src/pilot/config_idrac.py
```diff
@@ -248,6 +444,25 @@ def config_idrac_settings(drac_client, ip_service_tag, password, node):
...
 
 
+def _get_nic_config(self, resource, class_name,
+                    selection_expression, attr_cls):
+    filter_query = (
+        'select * '
+        'from %(class)s '
+        'where %(selection_expression)s') % {
+        'class': class_name, 'selection_expression': selection_expression}
+    doc = self.client.enumerate(resource, filter_query=filter_query)
```
cdearborn: I don't see how the above line could work since there is no client variable defined.

src/pilot/config_idrac.py
```diff
@@ -204,6 +207,199 @@ def configure_bios_nics_boot_settings(drac_client, ip_service_tag, pxe_nic_id):
...
+            raise exceptions.DRACOperationFailed(
+                drac_messages=('Colliding attributes %r' % (
+                    set(result) & set(attribs))))
+
+        result.update(attribs)
+
+    return result
+
+
+def get_nic_setting(nic_id, attribute_name):
```
cdearborn: It's ok to have this method, but it needs to be changed to call the upstream list_nic_settings(), and then find the given NIC setting in the resulting list.

src/pilot/config_idrac.py
```diff
@@ -172,17 +175,17 @@ def configure_bios_nics_boot_settings(drac_client, ip_service_tag, pxe_nic_id):
...
-            if not drac_client.is_nic_legacy_boot_protocol_pxe(nic_id):
-                result = drac_client.set_nic_legacy_boot_protocol_pxe(nic_id)
+            if not is_nic_legacy_boot_protocol_pxe(nic_id):
+                result = set_nic_legacy_boot_protocol_pxe(nic_id, drac_client)
         else:
             try:
-                if not drac_client.is_nic_legacy_boot_protocol_none(nic_id):
-                    result = drac_client.set_nic_legacy_boot_protocol_none(
-                        nic_id)
+                if not is_nic_legacy_boot_protocol_none(nic_id):
```
cdearborn: I think you will have to pass drac_client in on the line above.

src/pilot/config_idrac.py
```diff
@@ -172,17 +175,17 @@ def configure_bios_nics_boot_settings(drac_client, ip_service_tag, pxe_nic_id):
...
         # Compare the NIC IDs case insensitively. Assume ASCII strings.
         if nic_id.lower() == pxe_nic_id.lower():
-            provisioning_mac = nic.mac_address.lower()
+            provisioning_mac = nic.mac.lower()
 
             # This is the NIC we want to PXE boot, so set it to PXE if it's
             # not set to PXE already
-            if not drac_client.is_nic_legacy_boot_protocol_pxe(nic_id):
-                result = drac_client.set_nic_legacy_boot_protocol_pxe(nic_id)
+            if not is_nic_legacy_boot_protocol_pxe(nic_id):
```
cdearborn: I think you will have to pass drac_client in on the line above.

jainprasuk1996:
@cdearborn Please take a look at the last commit in this PR. The changes have been tested by @uday344 as per your suggestions.

src/pilot/config_idrac.py
```diff
@@ -208,6 +212,172 @@ def configure_bios_nics_boot_settings(drac_client, ip_service_tag, pxe_nic_id):
...
+    :raises: WSManInvalidResponse when receiving invalid response
+    :raises: DRACOperationFailed on error reported back by the iDRAC
+             interface
+    :raises: NotFound when no settings for NIC found
+    """
+    settings = drac_client.list_nic_settings(nic_id)
+
+    # Were no settings found?
+    if not settings:
+        raise exceptions.NotFound(
```
cdearborn: Our long term objective is to completely eliminate discover_nodes/dracclient.  I took a look and only nic.py used this exception, and it is the only exception declared in discover_nodes/dracclient/exceptions.  In the case above, if no settings were found, then the nic_id passed in must not have been found, which means that the nic_id was invalid.  As a result, let's change the above to raise a dracclient.exceptions.InvalidParameterValue exception and change all of the references to NotFound to InvalidParameterValue.  Finally, delete the discover_nodes/dracclient/exceptions.py file.  This will get us one step closer to being able to delete discover_nodes/dracclient completely.

PR #235, opened 2019-01-23
Fix for is_boss_controller method and constants file using Monkey Patch

Rachit7194:
@cdearborn @digambar15 @dpaterson  Guys, i have fix the issue with is_boss_controller method by properly updating it's return value and added missing class in constant.py file. 
Please review it

### DIFF
--- a/src/pilot/constants.patch
+++ b/src/pilot/constants.patch
@@ -1,11 +1,19 @@
---- constants.py	2018-09-19 04:11:26.000000000 -0500
-+++ constants_backup.py	2019-01-08 12:18:12.983836000 -0600
-@@ -20,6 +20,8 @@
- DEFAULT_WSMAN_SSL_ERROR_RETRIES = 3
- DEFAULT_WSMAN_SSL_ERROR_RETRY_DELAY_SEC = 0
+--- constants.py.orig	2019-01-23 14:21:10.077000000 +0000
++++ constants.py	2019-01-23 13:45:52.275000000 +0000
+@@ -75,3 +75,16 @@
+                 cls.reboot_forced_shutdown]
  
-+NOT_SUPPORTED_MSG = " operation is not supported on th"
+ 
++class RaidStatus(object):
++    """Enumeration of different volume types."""
 +
- # power states
- POWER_ON = 'POWER_ON'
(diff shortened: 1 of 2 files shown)


Rachit7194:
I have tested this monkey patch manually and its updates both the files raid.py and constants.py

src/pilot/constants.patch
```diff
@@ -1,11 +1,19 @@
...
-+++ constants_backup.py	2019-01-08 12:18:12.983836000 -0600
-@@ -20,6 +20,8 @@
- DEFAULT_WSMAN_SSL_ERROR_RETRIES = 3
- DEFAULT_WSMAN_SSL_ERROR_RETRY_DELAY_SEC = 0
+--- constants.py.orig	2019-01-23 14:21:10.077000000 +0000
++++ constants.py	2019-01-23 13:45:52.275000000 +0000
+@@ -75,3 +75,16 @@
+                 cls.reboot_forced_shutdown]
  
-+NOT_SUPPORTED_MSG = " operation is not supported on th"
```
cdearborn: The original constants.patch patched in the NOT_SUPPORTED_MSG constant, which the dracclient/resources/raid.py is dependent on.  The new patch does not patch that in.  We need to continue to patch in the NOT_SUPPORTED_MSG constant, or things will break.

cdearborn: Actually, I can see that the NOT_SUPPORTED_MSG constant is available in OSP13z4, so monkey patching it in is no longer needed.  Please ignore my above comment.

src/pilot/dracclient_raid.patch
```diff
@@ -1,205 +1,31 @@
...
-                     raise
+--- raid.py.orig	2019-01-23 15:57:50.655000000 +0000
++++ raid_backup.py	2019-01-23 14:44:59.068000000 +0000
+@@ -588,14 +588,26 @@
+         """
+         return raid_controller_fqdd.startswith('RAID.')
  
-         return is_jbod_capable
-+
-+    def is_raid_controller(self, raid_controller_fqdd):
```
cdearborn: Nice job eliminating these methods from the patch since they are now available in OSP13z4!

PR #238, opened 2019-01-31
CES-10074 - fix for possible osd config file contention if large amou…

gaelrehault:
fix for possible osd config file contention if large amount of storage nodes or slow connectivity to undercloud node.

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -926,7 +926,7 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
 
     # load the osd environment file
     osd_config_file = os.path.join(Constants.TEMPLATES, "ceph-osd-config.yaml")
-    stream = open(osd_config_file, 'r')
+    stream = open(osd_config_file, 'r+')
     while True:
         try:
             fcntl.flock(stream, fcntl.LOCK_EX | fcntl.LOCK_NB)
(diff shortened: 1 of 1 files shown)


src/pilot/assign_role.py
```diff
@@ -980,18 +994,39 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
     # do it the hard way to create something more user friendly
-    with open(osd_config_file + ".orig", 'r') as instream:
-        with open(osd_config_file, 'w') as outstream:
-            for line in instream:
-                if '{}' not in line:
-                    outstream.write(line)
+    instream = open(osd_config_file + ".orig", 'r')
+    while True:
+        try:
+            fcntl.flock(instream, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
cdearborn: I don't think it's necessary to lock on the .orig file.  This file is shipped as part of JetPack and it's never changed.  It should be ok to have multiple threads reading this file at once.

src/pilot/assign_role.py
```diff
@@ -980,18 +994,39 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
+    osd_config_str = json.dumps(node_data_lookup, sort_keys=True,
                                         indent=2, separators=(',', ': '))
-            for line in osd_config_str.split('\n'):
-                line = "    " + line + "\n"
-                outstream.write(line)
-
+    for line in osd_config_str.split('\n'):
+        line = "    " + line + "\n"
+        outstream.write(line)
+    fcntl.flock(instream, fcntl.LOCK_UN)
```
cdearborn: It seems like the unlocking and closing should be done in a finally block to ensure this happens even when an exception occurs.

src/pilot/assign_role.py
```diff
@@ -923,8 +926,19 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
+    while True:
+        try:
+            fcntl.flock(stream, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            break
+        except IOError as e:
+            if e.errno != errno.EAGAIN:
+                raise
+            else:
+                time.sleep(1)
+    current_osd_configs = yaml.load(stream)
```
cdearborn: I think there still is an issue here.  The scenario that I'm thinking of is 1 thread loads the current config and is then swapped out.  Another thread loads the current config, adds a new config to it and saves the file.  When the 1st thread starts back up, it will overwrite the file, losing the entry that the 2nd thread wrote to it.  To avoid this, I think we would have to lock the file for rw and do everything inside the lock in one shot instead of lock, read, release, lock, write.

gaelrehault:
@cdearborn  should be better now, i tested with > 50 storage nodes and didn t run into any issues

cdearborn:
Looks good!  It looks like there may be some pep8 warnings that need to be cleaned up, but we can do that later.

PR #239, opened 2019-02-04
Realtime raid jbod conversion

Rachit7194:
@cdearborn @digambar15 @dpaterson  Please review the code. Handle realtime support in change_physical_disk_state_wait() method.


src/pilot/assign_role.py
```diff
@@ -1366,13 +1366,21 @@ def change_physical_disk_state_wait(
...
+            realtime = drac_client.is_realtime_supported(controller_id)
+            realtime_controllers.append(realtime)
             job_id = drac_client.commit_pending_raid_changes(
-                controller_id, reboot=False, start_time=None)
+                controller_id, reboot=False, start_time=None,
+                realtime=realtime)
             job_ids.append(job_id)
 
+    # if all contollers support real time then is_reboot_required is false
+    if len(set(realtime_controllers)) == 1:
```
digambar15: if one node has three controllers, and is_realtime_supported returns [True, False, True], in this case set will return [False, True] and len of set is 2 so above condition will always fail

digambar15: this check looks good

cdearborn: The above check will be true if either all of the controllers support realtime or all of the controllers don't support realtime.  I think we need to change this so that it also makes sure that the value is true.

src/pilot/assign_role.py
```diff
@@ -1366,13 +1366,21 @@ def change_physical_disk_state_wait(
     change_state_result = drac_client.change_physical_disk_state(
         mode, controllers_to_physical_disk_ids)
 
+    realtime_controllers = []
     job_ids = []
     if change_state_result['commit_required_ids']:
         for controller_id in change_state_result['commit_required_ids']:
+            realtime = drac_client.is_realtime_supported(controller_id)
```
cdearborn: Calling is_realtime_supported in this loop will beat up the iDRAC a little because it will result in len(change_state_result['commit_required_ids']) list_controller calls to the iDRAC.  I think it would be better to just call list_controllers once just before this loop and build a dict that maps controller_id to a boolean that indicates if the controller supports realtime mode.  This would turn the above line into a lookup into the dict by controller_id.  This would make it so we make fewer calls to the iDRAC, which will speed things up, and not put unnecessary stress on the iDRAC.

src/pilot/assign_role.py
Rachit7194: Updated in new PR[#254](https://github.com/dsp-jetpack/JetPack/pull/254)

src/pilot/assign_role.py
Rachit7194: Updated in new PR[#254](https://github.com/dsp-jetpack/JetPack/pull/254)

Rachit7194:
Guys i have created a new PR with reference to this PR 
Duplicate of #[254](https://github.com/dsp-jetpack/JetPack/pull/254). So closing this PR.

PR #240, opened 2019-02-05
iDRAC not ready for discover_nodes

mpardhi23:
@cdearborn @digambar15 Please review the code. Invoke made using is_idrac_ready() method in discover_nodes module.

### DIFF
--- a/src/pilot/discover_nodes/discover_nodes.py
+++ b/src/pilot/discover_nodes/discover_nodes.py
@@ -447,20 +447,7 @@ def is_idrac(client):
     # This determines whether or not an IPv4 address is a WS-Man
     # endpoint and iDRAC.
     #
-    # The Dell Common Information Model's (DCIM) DCIM_iDRACCardView
-    # class is leveraged. That class is documented in the "iDRAC Card
-    # Profile"
-    # (http://en.community.dell.com/techcenter/extras/m/white_papers/20441091/download),
-    # which is part of the DCIM Extensions Library Profile Collection.
-    # The collection is available at
-    # http://en.community.dell.com/techcenter/systems-management/w/wiki/1906.dcim-library-profile.
-    #
-    # An instance of that class represents an iDRAC's information. This
-    # assumes that at most only one (1) instance can be obtained from a
-    # WS-Man service implementation that is an iDRAC. WS-Man's Enumerate
-    # operation is used to obtain it.
-    #
-    # since the GetRemoteAPIStatus call is vendor specific, if the
+    # Since the GetRemoteAPIStatus call is vendor specific, if the
     # server responds then that is sufficient to determine that it's an iDRAC.
     # Squelch a couple of chatty libraries.
     dracclient_wsman_logger = logging.getLogger('dracclient.wsman')
@@ -470,7 +457,7 @@ def is_idrac(client):
     requests_logger.disabled = True
 
     try:
-        return client.client.is_idrac_ready()
+        is_ready = client.client.is_idrac_ready()
     except dracclient.exceptions.WSManInvalidResponse as e:
         # Most likely the user credentials are unauthorized.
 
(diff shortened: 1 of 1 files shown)


src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -501,15 +501,8 @@ def is_idrac(client):
...
-    if doc is None:
-        return False
-
-    return dracclient.utils.get_wsman_resource_attr(
-        doc,
-        DCIM_iDRACCardView,
-        'DeviceDescription') == 'iDRAC'
-
+    
+    return is_ready
```
cdearborn: The server is an iDRAC if the is_idrac_ready call returns true or false.  As a result we don't want to return the value of is_ready, we just want to return a hard-coded true if no exception occurred.

src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -473,7 +473,7 @@ def is_idrac(client):
     requests_logger.disabled = True
 
     try:
-        doc = client.client.enumerate(DCIM_iDRACCardView)
+        is_ready = client.client.is_idrac_ready()
```
cdearborn: Can you change the comment for this method so that it no longer references DCIM_iDRACCardView?

You could replace it with a comment that just says that since the GetRemoteAPIStatus call is vendor specific, if the server responds to that then that is sufficient to determine that it's an iDRAC.

mpardhi23: Updated the comment and also changed the return value as suggested.

mpardhi23:
Tested the code changes with the suggested approach.
Case-1:  Ran is_idrac_ready.py successfully, when node is in life-cycle controller setup.
Case-2: Ran discover_nodes.py successfully, when node is in life-cycle controller setup.
Case-3: Ran config_idrac.py successfully, when node is out of life-cycle controller setup.

src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -460,11 +460,8 @@ def is_idrac(client):
     # WS-Man service implementation that is an iDRAC. WS-Man's Enumerate
     # operation is used to obtain it.
     #
-    # This returns true if the Enumerate operation succeeds, the
-    # response contains an DCIM_iDRACCardView instance, the instance has
-    # a 'DeviceDescription' property, and the value of that property is
-    # 'iDRAC'; otherwise, false is returned.
-
+    # since the GetRemoteAPIStatus call is vendor specific, if the
```
cdearborn: Can you remove lines 450-461 since they are no longer correct?

src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -473,7 +470,7 @@ def is_idrac(client):
     requests_logger.disabled = True
 
     try:
-        doc = client.client.enumerate(DCIM_iDRACCardView)
+        return client.client.is_idrac_ready()
```
cdearborn: We don't want to return a value indicating if the idrac is ready or not.  We want to return True if the is_idrac_ready call does not throw an exception regardless of if it returns True or False.

mpardhi23: Updated code in is_idrac() method for return value as suggested and also tested the code changes with the suggested approach.
Removed the lines which are referencing to DDCIM_iDRACCardView in comments.

src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -501,14 +485,8 @@ def is_idrac(client):
...
         dracclient_wsman_logger.disabled = False
-
-    if doc is None:
-        return False
-
-    return dracclient.utils.get_wsman_resource_attr(
-        doc,
-        DCIM_iDRACCardView,
-        'DeviceDescription') == 'iDRAC'
+    if is_ready in (True, False):
```
cdearborn: The above line is not needed.  If any exception occurs, it will be caught above at line 496.  If we made it this far in the code, then the server responded to the is_idrac_ready check, so we can just "return True" here.

mpardhi23: Removed extra if statement and updated the code returning "True" value.

PR #242, opened 2019-02-07
Add support for Glance Backend Option

rajinir:
Define the backend to use for Glance in the ini file.
For cinder
glance_backend=cinder
Set to rbd to use ceph as backend.
glance_backend=rbd
Set to file to use file-based storage for images.
glance_backend=file
And deploy overcloud
Note: cinder has to be configured to use the right backend
if cinder is choosen

### DIFF
--- a/src/deploy/osp_deployer/director.py
+++ b/src/deploy/osp_deployer/director.py
@@ -438,6 +438,7 @@ class Director(InfraHost):
         osd_dedicated_devices = "    dedicated_devices:\n"
         domain_param = "  CloudDomain:"
         rbd_backend_param = "  NovaEnableRbdBackend:"
+        glance_backend_param = "  GlanceBackend:"
         ceph_pool_default_size_param = "  CephPoolDefaultSize:"
         ceph_pool_default_size = 3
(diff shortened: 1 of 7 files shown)


heydog15:
Looks good to me.  Waiting on Gael's review.

gaelrehault:
"_Note: cinder has to be configured to use the right backend if cinder is choosen_"
Could you expand on that please ?

gaelrehault:
Could you update the xsp and csp profiles definition (json's) to enforce the allowed values for glance_backend (file, rbd or cinder)

rajinir:
> Could you update the xsp and csp profiles definition (json's) to enforce the allowed values for glance_backend (file, rbd or cinder)

Done. Good suggestion. Also added swift as another option.

rajinir:
> "_Note: cinder has to be configured to use the right backend if cinder is choosen_"
> Could you expand on that please ?

These are options for glance backend , if cinder is specified, then whatever default cinder-volume  backend configured will be used. Same with swift and if rbd is choosen will use ceph.

PR #245, opened 2019-02-08
This patch fixes an intermittent issue when configuring the UEFI PXE …

cdearborn:
…NIC (#244)

* This patch fixes an intermittent issue when configuring the UEFI PXE NIC

The issue is that the PxeDev1EnDis attribute is randomly set to disabled.
This causes the PxeDev1Interface attribute to be read-only, which in
turn causes python-dracclient to throw an exception because we're trying
to set a read-only attribute.

Note that this happens only sometime on the latest firmware.

The fix changes config_idracs as follows:
1. Set the boot mode in a job and commit it.
   This will cause an extra reboot for BIOS boot mode.
2. When the boot mode is UEFI, set PxeDev1EnDis in a job and commit it.
   This causes an extra reboot for UEFI boot mode.
3. After PxeDev1EnDis has been set, set PxeDev1Interface
   This change will be bundled in with the reboot for all the other jobs.

* Updated from review comments.

### DIFF
--- a/src/pilot/config_idrac.py
+++ b/src/pilot/config_idrac.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 
-# Copyright (c) 2017-2018 Dell Inc. or its subsidiaries.
+# Copyright (c) 2017-2019 Dell Inc. or its subsidiaries.
 #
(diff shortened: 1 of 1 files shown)


rajinir:
+1

PR #246, opened 2019-02-08
Fix for RAID cleaning timeout failure

cdearborn:
This patch doubles the RAID cleaning timeout as it was not long enough for
some servers.


heydog15:
Looks good.  Have done at least 3 installs now without seeing the cleaning issue again.

PR #249, opened 2019-02-12
Fix jbod empty logical disk 

digambar15:


digambar15:
> Great job hunting this issue down and fixing it! I took a close look, and because define_jbod_logical_disks is only called from this one place, I think it would be better to put the filtering logic inside that method so that if we ever add code that calls it in the future, we won't have to put the filtering logic in that caller too.
> 
> In the define_jbod_logical_disks method, we have the following statements:
> if jbod_logical_disk is not None:
> logical_disks.append(jbod_logical_disk)
> 
> I think if we just changed the "if" statement to the following, then that would be a slightly simpler fix that would ensure the method never returns an empty dict in the list:
> if jbod_logical_disk:
> logical_disks.append(jbod_logical_disk)

Agree

digambar15:
@cdearborn Updated code

PR #252, opened 2019-02-13
Handle tempest config if public network not avail

dpaterson:
If sanity test is not run prior to attempting to
configure tempest an excption would occur when the
tempest configuration script was called.  This patch
insures that the network the config script requires
is present, if it is not then the user is instructed
to run the sanity test to create the network and
to rerun the deployment with the --tempest_config_only
flag.

Also added --tempest_config_only flag to help user
recover from error where tempest.conf could not be
generated due to the sanity test failing or not
running at all, as executing it is currently an
option.

Fixes: CES-10320

### DIFF
--- a/src/deploy/osp_deployer/director.py
+++ b/src/deploy/osp_deployer/director.py
@@ -1520,13 +1520,24 @@ class Director(InfraHost):
             pass
 
     def configure_tempest(self):
-        logger.debug("configuring tempest")
+        logger.debug("Configuring tempest")
         setts = self.settings
+        external_sub_cmd = ("source ~/" + setts.overcloud_name + "rc;" +
+                            "openstack subnet list  | grep external_sub " +
+                            "| awk '{print $6;}'")
+        external_sub_guid = self.run_tty(external_sub_cmd)[0].rstrip()
+
+        if not external_sub_guid:
+            logger.error("Could not find public network, please run the " +
+                        "sanity test to create the appropriate networks " +
+                        "and re-run this script with " +
+                        "--tempest_config_only flag.")
+            pass
+
         cmds = [
-            'source ~/' + self.settings.overcloud_name + 'rc;'
-            "sudo ip route add " + self.settings.floating_ip_network +
-            " dev eth0",
-            'source ~/' + self.settings.overcloud_name + 'rc;' +
+            'source ~/' + setts.overcloud_name + 'rc;'
+            "sudo ip route add " + setts.floating_ip_network + " dev eth0",
+            'source ~/' + setts.overcloud_name + 'rc;' +
             'tempest init mytempest;cd mytempest;' +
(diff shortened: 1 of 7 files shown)


src/pilot/templates/manila-unity-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ManilaUnityBackendName: tripleo_manila_unity
+  ManilaUnityDriverHandlesShareServers: <manila_unity_driver_handles_share_servers> 
+  ManilaUnityNasLogin: <manila_unity_nas_login>
+  ManilaUnityNasPassword: <manila_unity_nas_password>
+  ManilaUnityNasServer: <manila_unity_nas_server>
+  ManilaUnityServerMetaPool: <manila_unity_server_meta_pool>
+  ManilaUnityShareDataPools: <manila_unity_share_data_pools>
+  ManilaUnityEthernetPorts: <manila_unity_ethernet_ports>
+  ManilaUnityEmcSslCertVerify: <manila_unity_ssl_cert_verify>
+  ManilaUnityEmcSslCertPath: <manila_unity_ssl_cert_path>
```
rajinir: This file is already merged into master

src/deploy/osp_deployer/director.py
```diff
@@ -415,7 +418,7 @@ def setup_templates(self):
 
         self.setup_networking()
         self.setup_dell_storage()
-        self.setup_manila() 
+        self.setup_manila()
```
rajinir: Manila changes are merged. Rebase?

src/deploy/osp_deployer/director.py
```diff
@@ -1520,41 +1523,60 @@ def run_sanity_test(self):
...
+                            "openstack subnet list  | grep external_sub " +
+                            "| awk '{print $6;}'")
+        external_sub_guid = self.run_tty(external_sub_cmd)[0].rstrip()
+
+        if not external_sub_guid:
+            logger.error("Could not find public network, please run the " +
+                         "sanity test to create the appropriate networks " +
+                         "and re-run this script with " +
+                         "--tempest_config_only flag.")
+            pass
```
cdearborn: Do we want to about in this case instead of attempting to run tempest, since we know it won't work?

src/deploy/osp_deployer/director.py
```diff
@@ -1520,41 +1523,60 @@ def run_sanity_test(self):
...
             self.run_tty(cmd)
 
     def run_tempest(self):
-        logger.debug("running tempest")
+        logger.debug("Running tempest")
+
+        if not self._is_tempest_conf():
+            self.configure_tempest()
+
+        os._exit(0)
```
cdearborn: There's definitely something wrong here.  We exit unconditionally, and there is a bunch of code following that won't get executed.

PR #253, opened 2019-02-14
Octavia support for JS 13.1

rashidxf:
Added a flag in settings.ini files to enable/disable Octavia load balancer at deployment time.
Also, provided an option for Users to disable Octavia generated certificates and keys for secure communication between Octavia containers and amphora load balancer and provide their own certificates and keys in settings.ini files.

### DIFF
--- a/src/deploy/osp_deployer/director.py
+++ b/src/deploy/osp_deployer/director.py
@@ -1062,9 +1062,10 @@ class Director(InfraHost):
             cmd += " --ovs_dpdk"
         if self.settings.enable_sriov is True:
             cmd += " --sriov"
-        # Node placement is required in an automated install. The index order
         if self.settings.dvr_enable is True:
             cmd += " --dvr_enable"
+        if self.settings.octavia_enable is True:
+            cmd += " --octavia_enable"
         # Node placement is required in an automated install.  The index order
         # of the nodes is the order in which they are defined in the
         # .properties file
--- a/src/deploy/osp_deployer/director.py
+++ b/src/deploy/osp_deployer/director.py
@@ -688,13 +688,13 @@ class Director(InfraHost):
                            self.templates_dir + "/cert_keys.yaml")
           self.run_tty('sed -i "s|OctaviaGenerateCerts:.*|' + 
             'OctaviaGenerateCerts: ' +
-            'False' + '|" ' + 
+            'false' + '|" ' + 
             str(octavia_yaml))
         
         if self.settings.octavia_user_certs_keys is False:
           self.run_tty('sed -i "s|OctaviaGenerateCerts:.*|' + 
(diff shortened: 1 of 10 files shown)


src/deploy/osp_deployer/director.py
```diff
@@ -678,10 +678,25 @@ def setup_net_envt(self):
...
         self.upload_file(self.settings.network_env_yaml,
                          network_yaml)
-
+                         
+        if self.settings.octavia_user_certs_keys is True:
+          self.upload_file(self.settings.certificate_keys_path,
+                           self.templates_dir + "/cert_keys.yaml")
+          self.run_tty('sed -i "s|OctaviaGenerateCerts:.*|' + 
+            'OctaviaGenerateCerts: ' +
+            'False' + '|" ' +
```
gaelrehault: can you make is lowercase (false) to match up with default template s case

src/deploy/osp_deployer/director.py
```diff
@@ -678,10 +678,25 @@ def setup_net_envt(self):
...
+                           self.templates_dir + "/cert_keys.yaml")
+          self.run_tty('sed -i "s|OctaviaGenerateCerts:.*|' + 
+            'OctaviaGenerateCerts: ' +
+            'False' + '|" ' + 
+            str(octavia_yaml))
+        
+        if self.settings.octavia_user_certs_keys is False:
+          self.run_tty('sed -i "s|OctaviaGenerateCerts:.*|' + 
+            'OctaviaGenerateCerts: ' +
+            'True' + '|" ' +
```
gaelrehault: can you make is lowercase (true) to match up with default template s case

src/deploy/osp_deployer/director.py
rashidxf: this has been fixed

src/deploy/osp_deployer/director.py
rashidxf: this has been fixed

src/pilot/deploy-overcloud.py
```diff
@@ -464,6 +472,14 @@ def main():
...
             env_opts += " -e ~/pilot/templates/neutron-ovs-dvr.yaml"
 
+        # The octavia.yaml must be included after the
+        # network-environment.yaml
+        if args.octavia_enable:
+            env_opts += " -e ~/pilot/templates/octavia.yaml"          
+
+        if args.octavia_user_certs_keys is True:
+            env_opts += " -e ~/pilot/templates/cert_keys.yaml"
+
```
gaelrehault: You probably only want to enable the cert yaml if octavia is enabled as well.

rashidxf: Yes you are right. This is fixed now.

PR #256, opened 2019-02-15
Fix for CES-10376: Sanity test is failing to ping vlan-aware VM

rashidxf:
Sanity was trying to ping the VM before it was active. Adding the delay of 20 gives the instance time to be active before sanity test can ping it.


heydog15:
Looks good.

PR #262, opened 2019-02-20
updating headers year to current

gaelrehault:


heydog15:
Looks good.  I didn't go through all 124 but did a sample check and it looks like you got them all covered.

PR #269, opened 2019-02-21
CES-10377

gaelrehault:
removes cryptography deprecation warnings


heydog15:
Looks good.

heydog15:
Manually tested on SAH and installed and cleared up issue.  A warning did come back stating there is a newer version PIP though that we might want to consider.

PR #270, opened 2019-02-21
cleaning up pep

gaelrehault:


src/pilot/deploy-overcloud.py
```diff
@@ -515,8 +516,10 @@ def main():
 
         if args.enable_dellsc:
             env_opts += " -e ~/pilot/templates/dell-cinder-backends.yaml"
-        
+
         if args.enable_unity:
+            env_opts += " -e ~/pilot/templates/dellemc-unity-cinder-" \
+                        "backends.yaml"
```
cdearborn: looks like the above line should be "backend.yaml", and the line below should be deleted.

gaelrehault: fixed, thanks

PR #272, opened 2019-02-22
Autocalculate pg_nums

cdearborn:
This patch changes the ceph configuration so that instead of reducing the replication count to fit hardcoded pg_nums and the number of OSDs in the cluster, it counts up the number of OSDs in the cluster, leaves the replication factor set to 3, and calculates what the pg_nums should be for each pool.  It then updates the dell-environment.yaml with these numbers.

### DIFF
--- a/src/deploy/osp_deployer/director.py
+++ b/src/deploy/osp_deployer/director.py
@@ -27,10 +27,17 @@ import subprocess
 import sys
 import tempfile
 import time
+import yaml
 
 logger = logging.getLogger("osp_deployer")
 
 exitFlag = 0
+
+# Ceph pools present in a default install
+HEAVY_POOLS = ['volumes', 'images', 'vms']
+OTHER_POOLS = ['rbd', '.rgw.root', 'default.rgw.control',
+               'default.rgw.data.root', 'default.rgw.gc', 'default.rgw.log',
+               'metrics', 'backups', '.rgw.buckets', 'default.rgw.users.uid']
 
 
 class Director(InfraHost):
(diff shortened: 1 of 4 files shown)


src/deploy/osp_deployer/settings/config.py
```diff
@@ -494,6 +496,8 @@ def __init__(self, settings_file):
             '/pilot/templates/dell-cinder-backends.yaml'
         self.dell_unity_cinder_yaml = self.foreman_configuration_scripts + \
             '/pilot/templates/dellemc-unity-cinder-backend.yaml'
+        self.ceph_osd_config_yaml = self.foreman_configuration_scripts + \
+            '/pilot/templates/ceph-osd-config.yaml'
```
pauldardeau: Did you intend to use CEPH_OSD_CONFIG_FILE (class variable) here?

src/deploy/osp_deployer/director.py
```diff
@@ -408,6 +414,76 @@ def setup_templates(self):
         self.setup_environment()
         self.setup_sanity_ini()
 
+    def clamp_min_pgs(self, num_pgs):
+        pg_options = [8192, 4096, 2048, 1024, 512, 256, 128, 64, 32, 16, 8, 4,
+                      2, 1]
+        pg_max = pg_options[0]
```
pauldardeau: pg_max variable not used.  i believe this is a remnant of earlier experimentations

src/deploy/osp_deployer/director.py
```diff
@@ -408,6 +414,76 @@ def setup_templates(self):
...
+                num_pgs = pg_options[option_index]
+                finding_pg_value = False
+            else:
+                option_index += 1
+
+        return num_pgs
+
+    def calc_pgs(self, num_osds, num_heavy_pools, num_other_pools):
+        replication_factor = 3
+        max_pgs = num_osds * 200 / replication_factor
```
pauldardeau: I'm puzzled seeing replication factor here.  How does that fit into calculating PGs?

src/deploy/osp_deployer/director.py
cdearborn: removed.

src/deploy/osp_deployer/settings/config.py
cdearborn: Oops!  Yes.  Fixed.

src/deploy/osp_deployer/director.py
cdearborn: I *think* that was in your original algorithm.  If you go to the following link and search for "Logic behind Suggested PG Count", there's a formula that talks about calculating the number of pgs that includes dividing by the replication factor: https://ceph.com/pgcalc/

pauldardeau: The relation between PGs value and replication is very loose/indirect.  That's why I was surprised to see it.  I'm ok with it being there.  The logic to clamp the values is solid and would be able to resolve weird cases involving non-standard replica values (e.g., 2 is now common for all NVMe).

pauldardeau:
LGTM

PR #279, opened 2019-03-04
 removed deprecated undercloud setting

gaelrehault:


cdearborn:
I just wanted to follow up on this.  I did a new director installation and verified that cleaning_network is being set to ctlplane in ironic.conf, so removing this is the right thing to do.

PR #283, opened 2019-03-06
Unity cinder and manila container deployment

rajinir:
RH Container Catalog Login access to registry.connect.redhat.com is required. Subscription login should also work


gaelrehault:
Looks ok to me, but we need more details as to what the credentials are and why they are required, maybe a pointer to RH doc on the subject as well. 
Also could we not use the same credentials as rhsm_user and password already in settings to simplify ?

rajinir:
Added more details about the container login

rajinir:
I think I agree. I removed the container catalog login. Originally, it wasn't clear. Thanks

PR #284, opened 2019-03-06
Handle tempest config if public network not avail 

dpaterson:
If sanity test is not run prior to attempting to
configure tempest an excption would occur when the
tempest configuration script was called. This patch
insures that the network the config script requires
is present, if it is not then the user is instructed
to run the sanity test to create the network and
to rerun the deployment with the --tempest_config_only
flag.

Added new mutually exclusive arguments:
--tempest_config_only - helps the user
recover from error where tempest.conf could not be
generated due to the sanity test failing or not
running at all, as executing it is currently an
option.  If tempest.conf already exists it is backed
up and a new one created.

--run_tempest_only - Just run Tempest.  If no tempest.conf
is found we try to generate one then run.

Lastly, added new mutually exclusive argument group so the
user won't be able to run with both of these new args at the
same time as it will lead to confusion.

Fixes: CES-10320


dpaterson:
OMG conflicts, sigh

src/deploy/osp_deployer/deployer.py
```diff
@@ -54,20 +55,42 @@ def get_settings():
...
 
 
 def run_tempest():
-    settings = get_settings()
-    if settings.run_tempest is True:
+    settings, args = get_settings()
+    if settings.run_tempest is True or args.run_tempest_only:
         logger.info("=== Running tempest ==")
         director_vm = Director()
+        is_tempest_conf = director_vm.is_tempest_conf()
```
cdearborn: It doesn't look to me like the above line could work because the method in the Director class is _is_tempest_conf().

src/deploy/osp_deployer/deployer.py
```diff
@@ -215,6 +225,8 @@ def deploy():
         director_vm.node_introspection()
         director_vm.update_sshd_conf()
         director_vm.assign_node_roles()
+        # Should we be calling function below?
+        # director_vm.revert_sshd_conf()?
         director_vm.revert_sshd_conf
```
gaelrehault: We need to temporary do this so we can run assign role against all the nodes concurrently, but yes i believe that restoring the config as it was once we're done is good practice.

dpaterson: So method should be called `director_vm.revert_sshd_conf()` correct?  Will patch in separate PR

PR #286, opened 2019-03-07
DellSc Config deployment refactoring

rajinir:
Use Tripleo environment file for deploying all cinder backends. This simplifies deployment and works on upgrades


rajinir:
This needs review after JS 13.1

PR #287, opened 2019-03-07
add checks for no OSDs

pauldardeau:


src/deploy/osp_deployer/director.py
```diff
@@ -553,7 +558,15 @@ def setup_environment(self):
                     logger.warning(
                         "Bad entry in osd_disks: {}".format(osd))
 
+        if osds_per_node < 1:
```
cdearborn: I don't think we can have this check (561-563) because it's perfectly ok to omit the OSD configuration from the .properties file if you have 14g storage nodes with BOSS+HBA330 since we auto-generate the OSD config in this case.

src/deploy/osp_deployer/director.py
```diff
@@ -553,7 +558,15 @@ def setup_environment(self):
...
+            logger.info("no OSDs available. exiting")
+            sys.exit(1)
+
         total_osds = self.calc_num_osds(osds_per_node)
+        if total_osds == 0:
+            logger.info("no OSDs available. exiting")
```
cdearborn: Could you beef up the logged message a little?  Maybe say something like: Either the OSD configuration is not specified in the .properties file, or the storage nodes have no available storage to dedicate to OSDs.  Or something along those lines so the user has an idea of how to correct the problem.

src/deploy/osp_deployer/director.py
pauldardeau: I removed it.

src/deploy/osp_deployer/director.py
pauldardeau: I expanded the message as suggested.

PR #289, opened 2019-03-08
Added the manila docker image url for unity

rajinir:


rajinir:
Deployed twice over the weekend to test and it deployed fine.

PR #293, opened 2019-03-13
Refactor tempest config

dpaterson:
When doing a full deployment instead of outputting an exception
when sanity test has NOT been run, add a check to see if the
subnet sanity test should have created exists prior to trying
to configure tempest.


heydog15:
Looks good.

jtwill98:
+1 LGTM -- merging

PR #296, opened 2019-03-14
Unity: Added support for Manila Share Types

rajinir:
Create Share types


heydog15:
Looks good.  Rajini said she deployed successfully with this.

PR #297, opened 2019-03-15
Remove this hack as OOB introspection is fixed

kpradip001:
The current fix is not needed anymore as upstream patch has already implemented the resolution.

For more details, please go through the below link

Fix OOB introspection to use pxe_enabled flag in idrac driver - https://review.openstack.org/#/c/641683/


Rachit7194:
You need to add a call in install-director.sh file to apply the patch

kpradip001:
> You need to add a call in [install-director.sh()](https://github.com/dsp-jetpack/JetPack/blob/master/src/pilot/install-director.sh) file to apply the patch

src/pilot/inspect.patch
```diff
@@ -0,0 +1,70 @@
+--- inspect.py	2019-03-18 04:26:06.117996201 -0400
```
rpioso: @cdearborn commented previously:

"This looks good, but a new version (z5) of RHOSP has just shipped. This version has the CPU count fix in it. As a result, you will need to regenerate the patch below with the latest shipped version of inspect.py, and that patch should not contain the CPU fix since the fix is already present."

While I have not put my eyes on z5, my impression is this entire file, `src/pilot/inspect.patch`, is not needed.

Have you had an opportunity to look at z5 and confirm it is not needed? If so, please remove it from this PR.

src/pilot/install-director.sh
```diff
@@ -266,6 +266,13 @@ apply_patch "sudo patch -b -s /usr/lib/python2.7/site-packages/dracclient/resour
 sudo rm -f /usr/lib/python2.7/site-packages/dracclient/resources/raid.pyc
 sudo rm -f /usr/lib/python2.7/site-packages/dracclient/resources/raid.pyo
 
+# This hacks in a patch to remove this hack as OOB introspection is fixed. 
+echo
+echo "## Patching Ironic iDRAC driver inspect.py.."
+apply_patch "sudo patch -b -s /usr/lib/python2.7/site-packages/ironic/drivers/modules/drac/inspect.py ${HOME}/pilot/inspect.patch"
+sudo rm -f /usr/lib/python2.7/site-packages/ironic/drivers/modules/drac/inspect.pyc
+sudo rm -f /usr/lib/python2.7/site-packages/ironic/drivers/modules/drac/inspect.pyo
+
```
rpioso: Please see my comment about `src/pilot/inspect.patch`. If that file is not needed, this change should be removed.

src/pilot/inspect.patch
digambar15: @rpioso  Agree

Here what we found after installing undercloud and overcloud. In director inspect.py has only calculate_cpus() method. I think we need this patch for current fix, when pxe_enable code will be available in RHOSP, then we can remove this change.

cdearborn: yes, this patch will still be needed for the pxe_enable fix.  The section of the patch that patches in the code for the cpu count fix should be removed since that code is already in OSP13z5

cdearborn: Ah, I see that you've already removed the patch for the cpu count fix.  Thanks!

rpioso: I put my eyes on z5 and agree that this patch is needed.

src/pilot/inspect.patch
```diff
@@ -0,0 +1,70 @@
...
++        pxe_dev_nics = []
++        pxe_params = ["PxeDev1EnDis", "PxeDev2EnDis",
++                      "PxeDev3EnDis", "PxeDev4EnDis"]
++        pxe_nics = ["PxeDev1Interface", "PxeDev2Interface",
++                    "PxeDev3Interface", "PxeDev4Interface"]
++
++        try:
++            bios_settings = client.list_bios_settings()
++        except drac_exceptions.BaseClientException as exc:
++            LOG.error(_LE('DRAC driver failed to list bios settings '
```
rpioso: Ditto

src/pilot/inspect.patch
```diff
@@ -0,0 +1,70 @@
...
++                if param in bios_settings and bios_settings[
++                        param].current_value == "Enabled":
++                    pxe_dev_nics.append(
++                        bios_settings[nic].current_value)
++        elif bios_settings["BootMode"].current_value == "Bios":
++            for nic in nics:
++                try:
++                    nic_cap = client.list_nic_settings(nic_id=nic.id)
++                except drac_exceptions.BaseClientException as exc:
++                    LOG.error(_LE('DRAC driver failed to list nic settings '
```
rpioso: Ditto

src/pilot/inspect.patch
```diff
@@ -0,0 +1,70 @@
+--- inspect.py	2019-03-18 04:26:06.117996201 -0400
++++ inspect_latest.py	2019-03-18 04:36:58.007653944 -0400
+@@ -119,10 +119,17 @@
+                       {'node_uuid': node.uuid, 'error': exc})
+             raise exception.HardwareInspectionFailure(error=exc)
+ 
++        pxe_dev_nics = self._get_pxe_dev_nics(client, nics, node)
++        if pxe_dev_nics is None:
++            LOG.warning(_LW('No PXE enabled NIC was found for node '
```
rpioso: The translators for log levels, such as `_LE()`, should not be present. They were removed during Queens and are not in the OSP13z5 version of `inspect.py`. You can see from the cherry pick that added PXE enabled to Queens that they are not there -- https://review.openstack.org/#/c/624388/2/ironic/drivers/modules/drac/inspect.py

src/pilot/install-director.sh
```diff
@@ -266,6 +266,13 @@ apply_patch "sudo patch -b -s /usr/lib/python2.7/site-packages/dracclient/resour
 sudo rm -f /usr/lib/python2.7/site-packages/dracclient/resources/raid.pyc
 sudo rm -f /usr/lib/python2.7/site-packages/dracclient/resources/raid.pyo
 
+# This hacks in a patch to remove this hack as OOB introspection is fixed.
```
rpioso: This comment is not clear to me. How about something like the following?

    # This hacks in a patch to out-of-band inspection that sets pxe_enabled on ports.

src/pilot/inspect.patch
```diff
@@ -0,0 +1,70 @@
...
++                        '%(node_uuid)s.'), {'node_uuid': node.uuid})
++        
+         for nic in nics:
+             try:
+                 port = objects.Port(task.context, address=nic.mac,
+-                                    node_id=node.id)
++                                    node_id=node.id,
++                                    pxe_enabled=(nic.id in pxe_dev_nics))
++
```
rpioso: nit: This newline is not in the Queens version that contains PXE enabled. Instead, it is below, between lines 19 and 20.

src/pilot/inspect.patch
```diff
@@ -0,0 +1,70 @@
...
++                        '%(node_uuid)s.'), {'node_uuid': node.uuid})
++        
+         for nic in nics:
+             try:
+                 port = objects.Port(task.context, address=nic.mac,
+-                                    node_id=node.id)
++                                    node_id=node.id,
++                                    pxe_enabled=(nic.id in pxe_dev_nics))
++                
+                 port.create()
```
rpioso: nit: The newline follows this line.

src/pilot/inspect.patch
```diff
@@ -1,25 +1,18 @@
...
-+        
+--- inspect.py	2019-03-27 09:53:02.035769407 +0000
++++ inspect_modified.py	2019-03-27 09:51:32.965097867 +0000
+@@ -122,8 +122,10 @@
          for nic in nics:
              try:
                  port = objects.Port(task.context, address=nic.mac,
 -                                    node_id=node.id)
 +                                    node_id=node.id,
 +                                    pxe_enabled=(nic.id in pxe_dev_nics))
```
Rachit7194: @kpradip001  Above lines(line no:7-11) should be remain in patch

cdearborn: I believe this patch is missing a section that sets:
pxe_dev_nics = self._get_pxe_dev_nics(...)

kpradip001:
> It looks like the call to _get_pxe_dev_nics was accidentally removed from inspect.patch

I have updated _get_pxe_dev_nics() in latest inspect.patch file, please review it.

src/pilot/inspect.patch
digambar15: It's updated in latest commit

src/pilot/inspect.patch
digambar15: Updated patch.

src/pilot/inspect.patch
digambar15: Yes, He did create patch based on Queens

src/pilot/install-director.sh
kpradip001: It's updated in latest commit

src/pilot/inspect.patch
kpradip001: removed this new line in latest commit and added below as you mentioned.

src/pilot/inspect.patch
kpradip001: updated in latest commit

src/pilot/inspect.patch
kpradip001: updated this in latest commit

kpradip001:
We have successfully tested this patch in full deployment.

PR #303, opened 2019-03-25
Unity: Support specific version of container image

rajinir:
Adding support to pull a specific version of the custom container for unity. The default is 'latest'. This helps to lock the version if the latest image is bad or not deployable.

### DIFF
--- a/src/deploy/osp_deployer/settings/sample_csp_profile.ini
+++ b/src/deploy/osp_deployer/settings/sample_csp_profile.ini
@@ -362,8 +362,15 @@ dellsc_ssn=CHANGEME
 dellsc_server_folder=cmpl_iscsi_servers
 dellsc_volume_folder=cmpl_iscsi_volumes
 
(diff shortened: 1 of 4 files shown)


rajinir:
I have deployed this and tested.

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -364,6 +364,7 @@ dellsc_volume_folder=cmpl_iscsi_volumes
 
 # Unity cinder parameters. 
 enable_unity_backend=false
+cinder_unity_container_version=latest
```
heydog15: Can we add a comment here?  Something like, "In most cases, "latest" will work fine.  If you need to change to a specific version, specify it here.

rajinir:
@Kurt please re-review

heydog15:
Looks good!!  Thanks for the changes.

PR #305, opened 2019-03-26
Made the following changes in sanity script

zxa13:
OVS Offload port creation for SR-IOV in Sanity script
    Deleted the ceph-osd-config.yaml.director


heydog15:
What is this PR for?  What is the Jira story/test case that this covers?

zxa13:
> What is this PR for? What is the Jira story/test case that this covers?

The SRIOV port creation command is different in case of OVS Offload enabled. This fixes the issue.  The Jira story is CES-10493 (https://jira.dellemc-community.org/browse/CES-10493) which is now in Testing on the Jira board.

heydog15:
Has this been tested?  Can you test this first with Offload disabled and make sure it works then enable it and test it again?

zxa13:
> Has this been tested? Can you test this first with Offload disabled and make sure it works then enable it and test it again?

Yes I have tested this and the offload port is created only when SR-IOV offload is enabled in the sanity.ini file.

PR #307, opened 2019-03-26
Cleanup

cdearborn:
This patch removes extraneous white space from the sample .ini files, syncs up the xsp and csp sample files so there are minimum differences between the two, and removes an OSD config file that appears to have been checked in accidentally.


src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -364,11 +365,11 @@ dellsc_volume_folder=cmpl_iscsi_volumes
...
 #
 # Unity cinder parameters.
-# 
+#
 enable_unity_backend=false
 # The Dell EMC Unity Container is published in the RH Container Catalog
-# registry.connect.redhat.com.
-# By default, the latest image of the RHOSP13 release is pulled.
+# registry.connect.redhat.com
+# By default the latest image of the RHOSP release is pulled.
```
heydog15: I think the , should still be after the "By default"

cdearborn: Fixed

heydog15:
Looks good.

PR #316, opened 2019-03-28
Arrange and add verbiage to the JetPack .ini files Backend section

rajinir:
There is a need to re-arrange and add verbiage to the JetPack .ini files Backend section
for multiple backends

### DIFF
--- a/src/deploy/osp_deployer/settings/sample_csp_profile.ini
+++ b/src/deploy/osp_deployer/settings/sample_csp_profile.ini
@@ -350,6 +350,53 @@ enable_fencing=true
 
 [Storage back-end Settings]
 
+# Storage Backends: Ceph, Dell EMC Unity, Dell EMC Storage Center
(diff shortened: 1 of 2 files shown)


heydog15:
Can we move enable_rbd_backend=, enable_rbd_nova_backend=, and glance_backend= settings to under the [Storage back-end Settings] section so all the storage settings are in one place?

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -350,6 +350,53 @@ enable_fencing=true
 
 [Storage back-end Settings]
```
heydog15: backend

rajinir: This will require a bunch of code changes

heydog15:
Looks good.

PR #328, opened 2019-04-08
Fix the sanity test for multiple instance creation

rajinir:


rajinir:
Ran this with 10 sanity instances on the mobile rack and it ran to success

PR #329, opened 2019-04-12
Manila Share Fixes

rajinir:
1. Create shares in tenant scope aka sanity
2. Use floating network to create shares
3. Allow access to the sanity VM
4. Mount the share to the VM


heydog15:
Looks good.  This tested successfully on your rack.

PR #332, opened 2019-04-22
This patch adds realtime support to RAID/JBOD disk conversion

cdearborn:
This patch modifies assign_role to use the new realtime support added to RAID/JBOD conversion in python-dracclient in this patch:
https://review.opendev.org/#/c/654608/


cdearborn:
Since Rajini has approved this, I'm going ahead and merging it.

PR #335, opened 2019-04-23
Fix is_raid_controller

cdearborn:
This change patches in the changes to python-dracclient to make is_raid_controller return True for BOSS cards.  It also makes the necessary
changes to assign_role to use this change.


rajinir:
+1

PR #349, opened 2019-05-21
Fix deploy with low OSD count

cdearborn:
This patch adds in specifying all of the Ceph OSD pool sizes, which prevents
unspecified pools from defaulting to 128 PGs.  This prevents the number of
PGs from exceeding the total available when you have a low OSD count.


heydog15:
Looks good.  You successfully deployed with this config as well.

cdearborn:
Yes.  Successful deployment and successful sanity test run.

heydog15:
Paul stated he is good with it as well.

PR #353, opened 2019-06-04
Monkey patch for reset_raid_config,clear_foreign_config and is_idrac_ready best practice

mpardhi23:
Created monkey patch of reset_raid_config, clear_foreign_config (RAID deletion enhancement) and is_ready best practice (Checking LCStatus instead of MessageID) for dracclient  client.py and raid.py .


cdearborn:
Looks good!

PR #356, opened 2019-06-17
Barbican support for JS 13.2

rashidxf:
Barbican support for JS 13.2 includes,

- Management of secrets in OpenStack
- Encryption of Cinder Volumes
- validation of Glance images


src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -271,6 +271,9 @@ sriov_vf_count=64
 # Set the following option to true/false for DVR
 dvr_enable=false
 
+# Set to true to enable Barbican
```
heydog15: Use the same verbiage as the option for DVR above and Octavia below.  Should say "Set the following option to true/false for Barbican

src/deploy/osp_deployer/settings/sample_xsp_profile.ini
```diff
@@ -271,6 +271,9 @@ sriov_vf_count=64
 # Set the following option to true/false for DVR
 dvr_enable=false
 
+# Set to true to enable Barbican
```
heydog15: Use the same verbiage as the option for DVR above and Octavia below.  Should say "Set the following option to true/false for Barbican

PR #357, opened 2019-06-18
Remove discover_nodes/python-dracclient

Rachit7194:
@cdearborn @rpioso  @digambar15


src/pilot/is_idrac_ready.py
```diff
@@ -66,7 +60,7 @@ def main():
         drac_user = node["pm_user"]
         drac_password = node["pm_password"]
 
-        drac_client = DRACClient(drac_ip, drac_user, drac_password)
+        drac_client = client.DRACClient(drac_ip, drac_user, drac_password)
```
cdearborn: I think you need to import client

src/pilot/assign_role.py
```diff
@@ -125,7 +121,7 @@ def parse_arguments():
                         metavar="ROLE")
     parser.add_argument("-f",
                         "--flavor-settings",
-                        default="~/pilot/flavors_settings.json",
+                        default="flavors_settings.json",
```
cdearborn: I don't think we should be changing the defaults or behavior for flavor_settings.json as part of this patch

Rachit7194: Okay. Replace as is

Rachit7194:
Hi Chris, deployment on R7 stamp completed with this change. Now all the testing completed with this PR.

PR #359, opened 2019-06-21
Take lifecycle controller out of recovery mode

mpardhi23:
modified config_idrac.py in JetStream calling directly into python-dracclient to check if a node is in recovery mode and take it out of recovery mode if necessary.
And also created monkey patch for client.py, job.py, utils.py and lifecycle_controller.py

### DIFF
--- a/src/pilot/config_idrac.py
+++ b/src/pilot/config_idrac.py
@@ -541,7 +541,7 @@ def reset_idrac(drac_client, ip_service_tag):
     drac_client.reset_idrac(wait=True)
 
 
-def get_lifecycle_out_of_recovery(drac_client, ip_service_tag):
+def take_lifecycle_out_of_recovery(drac_client, ip_service_tag):
     LOG.info("Setting lifecycle settings on {}".format(ip_service_tag))
     settings = {'Lifecycle Controller State': 'Enabled'}
     drac_client.set_lifecycle_settings(settings)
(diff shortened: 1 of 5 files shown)


src/pilot/config_idrac.py
```diff
@@ -541,6 +541,22 @@ def reset_idrac(drac_client, ip_service_tag):
...
+    drac_client.set_lifecycle_settings(settings)
+
+    LOG.info("Waiting for lifecycle controller to get out of"
+             "recovery mode")
+    job_id = drac_client.commit_pending_lifecycle_changes(reboot=False)
+    job_succeeded = wait_for_jobs_to_complete(
+        [job_id], drac_client, ip_service_tag)
+
+    if not job_succeeded:
+        raise RuntimeError("An error occurred while configuring "
```
cdearborn: Would you mind changing the error to: "An error occurred while taking the iDRAC on xxx out of recovery mode"

mpardhi23: Sure Chris. I will change the error as suggested.
Thank you!

mpardhi23:
Updated error message.

src/pilot/config_idrac.py
```diff
@@ -541,6 +541,22 @@ def reset_idrac(drac_client, ip_service_tag):
     drac_client.reset_idrac(wait=True)
 
 
+def get_lifecycle_out_of_recovery(drac_client, ip_service_tag):
```
cdearborn: Would you mind changing "get_lifecycle_out_of_recovery" to "take_lifecycle_out_of_recovery"?

mpardhi23: I have changed the method name to "take_lifecycle_out_of_recovery".

mpardhi23:
Thanks Chris!

PR #361, opened 2019-06-25
Remove get nodes import

dpaterson:
Pretty sure we don't need this import as we have get_nodes function in this module already:

from update_ssh_config import get_nodes

See line 65:
def get_nodes(ironic_client):
    return ironic_client.node.list(detail=True)


gaelrehault:
Looks like something went wrong with your PR, you have x41 file changed ?

jtwill98:
I think this PR needs closed and a new PR created with the fix Dave wants.   This is way too many files for such a simple change.

PR #367, opened 2019-07-11
Ironic patch update

cdearborn:
These changes remove and update patches for ironic code that has arrived in OSP from upstream.


heydog15:
Looks good to me.

src/pilot/dracclient_raid.patch
```diff
@@ -1,5 +1,5 @@
---- /usr/lib/python2.7/site-packages/dracclient/resources/raid.py	2019-05-09 15:37:11.450522239 +0000
-+++ raid.py	2019-05-09 15:37:04.219391200 +0000
+--- /usr/lib/python2.7/site-packages/dracclient/resources/raid.py	2019-07-11 15:22:53.819829111 +0000
```
jtwill98: I believe you can strip the date and time stamps from the file 
i.e.   delete  "2019-05-09 15:37:04.219391200 +0000" & "2019-07-11 15:22:53.819829111 +0000" and the patch file will still work.   Not a concern here, just an FYI.

src/pilot/install-director.sh
```diff
@@ -273,20 +273,6 @@ echo "## Updating .bash_profile..."
 echo "source ~/stackrc" >> ~/.bash_profile
 echo "## Done."
 
-# This hacks in a patch to allow realtime RAID creation.
```
jtwill98: In the future, to avoid the issue with patches it may be best to have a test to determine if upstream code contains the patch to ship attempting to apply the patch.  I did this in the pass.  The problem doing this is you forget to remove the patch code :-(

dpaterson: +1

src/pilot/dracclient_raid.patch
cdearborn: k - I just do a "diff -Naur ... > blah.patch"

src/pilot/install-director.sh
cdearborn: yeah, was thinking about that.  unfortunately a single .patch file may contain multiple fixes, so you may have some parts of the patch that merge ok, and some that don't, so you can't just skip a whole file if it doesn't apply cleanly.  will give it some thought

gaelrehault: On top of multiple fixes in a patch .. there is also the case where the implementation upstream of a patch would differ, even slightly, from ours , that s usually the case and the reason we dont assume it made it upstream/ can t exactly diff that.

PR #368, opened 2019-07-12
updated wsman patch

gaelrehault:


heydog15:
Gael tested and the undercloud now installs.

PR #370, opened 2019-07-17
Fix PXE booting issues during deployment

cdearborn:
This patch monkey patches in the fix for setting pxe_enabled on ironic ports
during OOB introspection.


heydog15:
Testing the fix now on the Mobile Rack.

heydog15:
Tested this fix on both Chris's stamp and the Mobile Rack.  All nodes pxe booted as expected.

PR #375, opened 2019-07-24
Unity Container Version Lock Support

rajinir:


rajinir:
Pep8 errors are not related to this patch and they are already fixed in the other pending patch

src/deploy/osp_deployer/director.py
```diff
@@ -161,6 +161,19 @@ def upload_update_conf_files(self):
             dest_file = self.home_dir + yaml
             self.upload_file(source_file, dest_file)
 
+            unity_lock = "/unity_container_lock.ini"
```
jtwill98: As a developer that has ported code to many operating systems, I'm not a fan of this as the directory separator "/" is hard coded to the file name which makes porting a nightmare.  I'm only going to comment here as an FYI.

src/deploy/osp_deployer/director.py
```diff
@@ -161,6 +161,19 @@ def upload_update_conf_files(self):
...
+            unity_lock_file = self.settings.lock_files_dir + unity_lock
+            if self.settings.enable_unity_backend is True:
+               cmd = "grep unity_cinder__container_version " + unity_lock_file +
+                     " | awk -F '=' '{print $2}'"
+               self.settings.cinder_unity_container_version = self.run_tty(cmd)
+            if self.settings.enable_unity_manila_backend is True:
+               cmd = "grep unity_manila_container_version " + unity_lock_file +
```
gaelrehault: Shouldn t this be _**manila_unity_container_version**_

src/deploy/osp_deployer/director.py
```diff
@@ -161,6 +161,19 @@ def upload_update_conf_files(self):
             dest_file = self.home_dir + yaml
             self.upload_file(source_file, dest_file)
 
+            unity_lock = "/unity_container_lock.ini"
+            unity_lock_file = self.settings.lock_files_dir + unity_lock
+            if self.settings.enable_unity_backend is True:
+               cmd = "grep unity_cinder__container_version " + unity_lock_file +
```
gaelrehault: Shouldn t this be grep **_cinder_unity_container_version_**

gaelrehault:
Besides the errors above that wouldn t apply the versions properly, I dont see any logic in there to apply those versions based on the value of the ini _if_ version locked is enabled. This looks to me that those values would always be picked up, in which case they could just be set in the .ini . 
I think this PR should be reverted and the issues adressed.

src/deploy/osp_deployer/director.py
rajinir: You are right, I have to fix this. Good catch

PR #376, opened 2019-07-25
Fixed HugePages low memory issue

zxa13:
Increased OS and Kernel memory when HugePages are enabled on compute nodes.


heydog15:
+2

PR #377, opened 2019-07-25
Changing default Introspection to OOB in CSP Profile

zxa13:
• The default introspection method is changed to out-of-band in CSP profile.
• The default option will perform OOB introspection on all nodes except when NUMA, HugePages or OVS-DPDK are enabled. In that case, computes nodes will switch to in-band introspection automatically. (A note is added for this in .ini file)
• When deploying with in-band introspection set to true in settings.ini file, all nodes will use in-band introspection regardless of any NFV feature enabled or not.


zxa13:
@cdearborn kindly review and merge. Thanks

src/deploy/osp_deployer/director.py
```diff
@@ -276,6 +276,17 @@ def node_discovery(self):
...
                 " to the number of nodes defined in .properties file")
+        if setts.numa_enable is True or setts.hpg_enable is True:
+            grep_data = self.run_tty("grep -n pm_addr ~/instackenv.json")[0].rstrip()
+            instack_nodes = list(grep_data.split(","))
+            compute_nodes = self.settings.compute_nodes
+            for compute_node in compute_nodes:
+                for instack_node in instack_nodes:
+                    if instack_node.find(compute_node.idrac_ip) > 0:
+                        extracted_line = int(instack_node.split(':')[0])
+                        cmd = "sed -i "+""+str(extracted_line+2)+"s/pxe_drac/pxe_ipmitool/"+" ~/instackenv.json"
```
cdearborn: The above line is changing the Ironic driver from the iDRAC driver to the IPMI driver.  While it is true that a side affect of this is that in-band introspection will be run on the node, this change will cause other major problems, like:
- We won't do RAID configuration of the node
- We won't do BIOS configuration of the node
This is because the IPMI driver doesn't support either RAID or BIOS configuration.
Because of this, we should always use the Ironic iDRAC (pxe_drac) driver and never use the IPMI driver.

cdearborn:
Since this is really an optimization, and I think there's more work that can be done today to complete this, I would propose that we close this PR and revisit this for the next release.

PR #380, opened 2019-07-26
Few fixes for dellsc and pep8 errors

rajinir:


rajinir:
I had these changes in my branch which didn't get amended properly. This change has been tested and deployed correctly

PR #382, opened 2019-07-28
moved recovery code from config_idrac to discover_node

mpardhi23:
@cdearborn @digambar15 
I have removed code from config_idrac.py and moved it to discover_node.py.
And I have also tested recovery_mode code changes in discover_node by running full deployment .
Instack.json is created properly and also able to take node out of recovery.

### DIFF
--- a/src/pilot/discover_nodes/discover_nodes.py
+++ b/src/pilot/discover_nodes/discover_nodes.py
@@ -512,7 +512,7 @@ def is_idrac(drac_client):
 
             if not job_succeeded:
                 raise RuntimeError("An error occurred while taking the iDRAC"
-                                   " on {} out of recovery mode")
+                                   " out of recovery mode")
         else:
             LOG.info("IP address is an iDRAC and the iDRAC is ready")
     except dracclient.exceptions.WSManInvalidResponse as e:
--- a/src/pilot/discover_nodes/discover_nodes.py
+++ b/src/pilot/discover_nodes/discover_nodes.py
@@ -456,7 +456,15 @@ def is_idrac(drac_client):
     requests_logger.disabled = True
 
     try:
-        drac_client.client.is_idrac_ready()
+        if drac_client.is_lifecycle_in_recovery():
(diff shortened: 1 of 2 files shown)


src/pilot/discover_nodes/discover_nodes.py
```diff
@@ -456,7 +456,15 @@ def is_idrac(drac_client):
...
 
     try:
-        drac_client.client.is_idrac_ready()
+        if drac_client.is_lifecycle_in_recovery():
+            settings = {'Lifecycle Controller State': 'Enabled'}
+            drac_client.set_lifecycle_settings(settings)
+
+            LOG.info("Waiting for lifecycle controller to get out of "
+                     "recovery mode")
+            drac_client.commit_pending_lifecycle_changes(reboot=False)
```
cdearborn: Don't we still need to wait for the job to complete before returning?

mpardhi23: No, we don't need to wait for job completion. whenever create_config_job() method invoked and create job it will get complete in few seconds.
If you want me to add wait_for_jobs_to_complete for safer side then I need to add implementation of wait_for_jobs_to_complete() method in discover_nodes.py

cdearborn: Well, I think the only reason this is working is because the next thing we do is get the model, which will wait until the iDRAC is ready.  I think I would rather explicitly wait until the job completes.

mpardhi23: Sure Chris! I will add wait_for_job_completion call.

cdearborn:
One of the goals of the manual install version of JetPack is to simplify the deployment of OpenStack as much as possible.  I'm opposed to adding another script since this would increase the number of steps that a user would have to take to deploy.  Taking a lifecycle controller out of recovery mode is not a destructive operation, and it doesn't even require a reboot of the server.  As a result, I believe that discover_nodes is the correct place for it.

PR #383, opened 2019-07-29
OVS-DPDK new network configurations

WajeehaHamid:
Included the OVS-DPDK network configurations for the following cases:

- OVS-DPDK 6 port
This will enable 2 port OVS-DPDK when a separate provisioning interface is not used 
- OVS-DPDK 8 port
This will enable 4 port OVS-DPDK when a separate provisioning interface is not used 
- OVS-DPDK with SRIOV 8 port
This will enable 2 port OVS-DPDK and 2 port SRIOV when a separate provisioning interface is not used


zxa13:
> This looks good, but there are some conflicts that need to be resolved.

I have resolved the commits. Kindly review them.

PR #384, opened 2019-07-29
Sanity script changes for SRIOV Offload

WajeehaHamid:


src/pilot/deployment-validation/sanity_test.sh
```diff
@@ -407,8 +408,12 @@ setup_glance(){
 sriov_port_creation(){
   #Creating ports in the tenant scope
   set_tenant_scope
+  if [ "$SMART_NIC_ENABLED" != false ];then
```
jtwill98: FYI:   I see and understand why you tested NOT false condition, however I prefer testing True, False and flagging other entries (i.e "Enabled", "1", etc. ) as invalid entries.

WajeehaHamid: Noted, will have to work on all NOT false conditions used in this script in next release.

PR #387, opened 2019-07-31
Support for BOSS+PERC in 14G storage nodes

digambar15:
Following features will be implemented as part of this PR -

1. PERC H740P is supported and able to create OSD's successfully.
2. No more need to add osd_disks in .properties file as ceph_osd_config.yaml is generating as expected and Ceph is also picking up all the devices successfully as part of OSD creation.
3. Two controllers (BOSS + PERC) are supported in 14G Storage nodes.

Note - Current testing is done only against PERC H740P controller. Soon we will perform testing both the controllers.

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -246,7 +246,7 @@ def get_drac_client(node_definition_filename, node):
 
 
 def define_target_raid_config(role, drac_client):
-    raid_controller_ids = get_raid_controller_id(drac_client)
+    raid_controller_ids = get_raid_controller_ids(drac_client)
 
     if not raid_controller_ids:
         LOG.critical("Found no RAID controller")
@@ -271,7 +271,7 @@ def define_target_raid_config(role, drac_client):
         'logical_disks': logical_disks} if logical_disks is not None else None
 
 
-def get_raid_controller_id(drac_client):
+def get_raid_controller_ids(drac_client):
     disk_ctrls = drac_client.list_raid_controllers()
 
     raid_controller_ids = []
@@ -386,42 +386,30 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
 
 
 def define_storage_logical_disks(drac_client, raid_controller_ids):
-    boss_controller = None
-    raid_controller = None
+    raid_controllers = []
     all_physical_disks = drac_client.list_physical_disks()
 
     if len(raid_controller_ids) > 1:
         for cntrl in raid_controller_ids:
-            if drac_client.is_boss_controller(cntrl):
-                boss_controller = cntrl
-            else:
-                raid_controller = cntrl
+            raid_controllers.append(cntrl)
     else:
-        if drac_client.is_boss_controller(raid_controller_ids[0]):
-            boss_controller = raid_controller_ids[0]
-        else:
-            raid_controller = raid_controller_ids[0]
+        raid_controllers.append(raid_controller_ids[0])
 
     # Get the drives controlled by the RAID controller
-    raid_cntlr_physical_disks = []
-    boss_cntrl_disks = []
+    raid_cntlr_physical_disks = {}
     for disk in all_physical_disks:
-        if disk.controller == boss_controller:
-            boss_cntrl_disks.append(disk)
-        elif disk.controller == raid_controller:
-            raid_cntlr_physical_disks.append(disk)
+        if disk.controller in raid_controllers:
+            if disk.controller in raid_cntlr_physical_disks:
+                raid_cntlr_physical_disks[disk.controller].append(disk)
+            else:
+                raid_cntlr_physical_disks[disk.controller] = [disk]
 
     # Make sure we have enough drives attached to the RAID controller to create
     # a RAID1
-
-    if boss_cntrl_disks:
-        num_raid_cntlr_physical_disks = len(boss_cntrl_disks)
-    else:
-        num_raid_cntlr_physical_disks = len(raid_cntlr_physical_disks)
-    if num_raid_cntlr_physical_disks < 2:
+    if all(len(raid_cntlr_physical_disks[cntrl]) < 2
+            for cntrl in raid_controllers):
         LOG.critical(
-            "Cannot configure RAID 1 with only {} drives; need at least two "
-            "drives".format(num_raid_cntlr_physical_disks))
+            "For configuring RAID 1 needs at least two drives")
         return None
 
     # Make sure we have at least one drive for Ceph OSD/journals
@@ -431,12 +419,17 @@ def define_storage_logical_disks(drac_client, raid_controller_ids):
             "configuration")
         return None
 
-    if boss_controller:
-        os_logical_disk = define_storage_operating_system_logical_disk(
-            boss_cntrl_disks, drac_client, boss_controller)
+    if len(raid_controllers) > 1:
+        boss_controller = [cntrl for cntrl in raid_controllers
+                           if drac_client.is_boss_controller(cntrl)][0]
+        if boss_controller:
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[boss_controller],
+                drac_client, boss_controller)
     else:
         os_logical_disk = define_storage_operating_system_logical_disk(
-            raid_cntlr_physical_disks, drac_client, raid_controller)
+            raid_cntlr_physical_disks[raid_controllers[0]],
+            drac_client, raid_controllers[0])
 
     if os_logical_disk is None:
         return None
@@ -446,19 +439,19 @@ def define_storage_logical_disks(drac_client, raid_controller_ids):
     # The ironic RAID 'physical_disks' property is optional. While it is
     # presently used by this script, it is envisioned that it will not
     # be in the future.
+    remaining_physical_disks = []
     if 'physical_disks' in os_logical_disk:
         os_physical_disk_names = os_logical_disk['physical_disks']
-        remaining_physical_disks = [d for d in raid_cntlr_physical_disks
-                                    if d.id not in os_physical_disk_names]
-    else:
-        remaining_physical_disks = raid_cntlr_physical_disks
+        for cntrl, disks in raid_cntlr_physical_disks.items():
+            _ = [remaining_physical_disks.append(disk) for disk in disks
+                 if disk.id not in os_physical_disk_names]
 
     # Define JBOD logical disks with the remaining physical disks.
     #
     # A successful call returns a list, which may be empty; otherwise,
     # None is returned.
     logical_disks = [os_logical_disk]
-    if raid_controller:
+    for raid_controller in raid_controllers:
         jbod_capable = drac_client.is_jbod_capable(raid_controller)
         jbod_logical_disks = define_jbod_logical_disks(
             drac_client, remaining_physical_disks, raid_controller,
@@ -468,7 +461,7 @@ def define_storage_logical_disks(drac_client, raid_controller_ids):
             return None
 
         logical_disks.extend(jbod_logical_disks)
-
+    LOG.debug("logical_disks: {}".format(logical_disks))
     return logical_disks
 
 
@@ -739,7 +732,7 @@ def configure_raid(ironic_client, node_uuid, role, os_volume_size_gb,
     when RAID configuration failed and return False. Further testing
     should uncover interesting error conditions.'''
 
-    if get_raid_controller_id(drac_client) is None:
+    if get_raid_controller_ids(drac_client) is None:
         LOG.warning("No RAID controller is present.  Skipping RAID "
                     "configuration")
         return True
@@ -1033,42 +1026,58 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
 def get_drives(drac_client):
     spinners = []
     ssds = []
-    physical_disks = drac_client.list_physical_disks()
     virtual_disks = drac_client.list_virtual_disks()
-    raid_disks = [vd.physical_disks for vd in virtual_disks
-                  if vd.raid_level == '1'][0]
+
+    raid0_disks = [vd for vd in virtual_disks if vd.raid_level != '1']
+
+    raid1_disks = [vd.physical_disks for vd in virtual_disks
+                   if vd.raid_level == '1'][0]
 
     # Getting all physical disks of raid_disks except RAID1 disks
-    raid_physical_disks = [pd for pd in physical_disks if pd.id not in raid_disks]
+    physical_disks_of_raid = [pd for pd in drac_client.list_physical_disks()
+                              if pd.id not in raid1_disks]
 
-    for physical_disk in raid_physical_disks:
-        # Eliminate physical disks that in a state other than non-RAID
-        if physical_disk.raid_status != "non-RAID" and \
-                physical_disk.raid_status == "failed":
-            LOG.info("Skipping disk {id}, because it has a RAID status of "
-                     "{raid_status}".format(
-                         id=physical_disk.id,
-                         raid_status=physical_disk.raid_status))
-            continue
+    for virtual_disk in raid0_disks:
+        phy_disk = [pd for pd in physical_disks_of_raid
+                    if virtual_disk.physical_disks[0] == pd.id][0]
 
-        # Eliminate physical disks that have an error status
-        if physical_disk.status == 'error':
-            LOG.warning("Not using disk {id}, because it has a status of "
-                        "{status}".format(id=physical_disk.id,
-                                          status=physical_disk.status))
-            continue
-
-        # Go ahead and use any physical drive that's not in an error state, but
-        # issue a warning if it's not in the ok or unknown state
-        if physical_disk.status != 'ok' and physical_disk.status != 'unknown':
-            LOG.warning("Using disk {id}, but it has a status of \""
-                        "{status}\"".format(id=physical_disk.id,
-                                            status=physical_disk.status))
-
-        if physical_disk.media_type == "hdd":
-            spinners.append(physical_disk)
+        if phy_disk.media_type == 'hdd':
+            spinners.append(virtual_disk)
         else:
-            ssds.append(physical_disk)
+            ssds.append(virtual_disk)
+
+        physical_disks_of_raid.remove(phy_disk)
+
+    if physical_disks_of_raid:
+        for physical_disk in physical_disks_of_raid:
+            # Eliminate physical disks that in a state other than non-RAID
+            if physical_disk.raid_status != "non-RAID" and \
+                    physical_disk.raid_status == "failed":
+                LOG.info("Skipping disk {id}, because it has a RAID status of "
+                         "{raid_status}".format(
+                             id=physical_disk.id,
+                             raid_status=physical_disk.raid_status))
+                continue
+
+            # Eliminate physical disks that have an error status
+            if physical_disk.status == 'error':
+                LOG.warning("Not using disk {id}, because it has a status of "
+                            "{status}".format(id=physical_disk.id,
+                                              status=physical_disk.status))
+                continue
+
+            # Go ahead and use any physical drive that's not in an error
+            # state, but issue a warning if it's not in the ok or unknown state
+            if physical_disk.status != 'ok' and \
+                    physical_disk.status != 'unknown':
+                LOG.warning("Using disk {id}, but it has a status of \""
+                            "{status}\"".format(id=physical_disk.id,
+                                                status=physical_disk.status))
+
+            if physical_disk.media_type == "hdd":
+                spinners.append(physical_disk)
+            else:
+                ssds.append(physical_disk)
 
     return spinners, ssds
 
@@ -1118,7 +1127,10 @@ def generate_osd_config_with_journals(controllers, osd_drives, ssds):
 
 
 def get_by_path_device_name(physical_disk, controllers):
-    disk_index = physical_disk.description.split(" ")[1]
+    if physical_disk.description.startswith("Virtual Disk"):
+        disk_index = physical_disk.description.split(" ")[2]
+    else:
+        disk_index = physical_disk.description.split(" ")[1]
     for controller in controllers:
         pci_bus_number = get_pci_bus_number(controller)
         if physical_disk.controller == controller.id and \
@@ -1144,7 +1156,7 @@ def get_by_path_device_name(physical_disk, controllers):
 
 def get_pci_bus_number(controller):
     if controller.model.startswith("PERC H730") and \
-           len(controller.bus.lower()) == 1:
+       len(controller.bus.lower()) == 1:
         pci_bus_number = '0' + controller.bus.lower()
     else:
         pci_bus_number = controller.bus.lower()
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -279,8 +279,6 @@ def get_raid_controller_id(drac_client):
     for cnt in disk_ctrls:
         if drac_client.is_raid_controller(cnt.id):
             raid_controller_ids.append(cnt.id)
-        elif drac_client.is_boss_controller(cnt.id):
-            raid_controller_ids.insert(0, cnt.id)
 
     return raid_controller_ids
 
@@ -459,18 +457,17 @@ def define_storage_logical_disks(drac_client, raid_controller_ids):
     #
     # A successful call returns a list, which may be empty; otherwise,
     # None is returned.
-    if not raid_controller:
-        raid_controller = boss_controller
-    jbod_capable = drac_client.is_jbod_capable(raid_controller)
-    jbod_logical_disks = define_jbod_logical_disks(
-        drac_client, remaining_physical_disks, raid_controller,
-        jbod_capable)
-
-    if jbod_logical_disks is None:
-        return None
-
     logical_disks = [os_logical_disk]
-    logical_disks.extend(jbod_logical_disks)
+    if raid_controller:
+        jbod_capable = drac_client.is_jbod_capable(raid_controller)
+        jbod_logical_disks = define_jbod_logical_disks(
+            drac_client, remaining_physical_disks, raid_controller,
+            jbod_capable)
+
+        if jbod_logical_disks is None:
+            return None
+
+        logical_disks.extend(jbod_logical_disks)
 
     return logical_disks
 
@@ -1123,25 +1120,30 @@ def generate_osd_config_with_journals(controllers, osd_drives, ssds):
 def get_by_path_device_name(physical_disk, controllers):
     disk_index = physical_disk.description.split(" ")[1]
     for controller in controllers:
+        pci_bus_number = get_pci_bus_number(controller)
         if physical_disk.controller == controller.id and \
                 controller.model.startswith("PERC H740P"):
             return ('/dev/disk/by-path/pci-0000:'
                     '{pci_bus_number}:00.0-scsi-0:2:{disk_index}:0').format(
-                pci_bus_number=controller.bus.lower(),
+                pci_bus_number=pci_bus_number,
                 disk_index=disk_index)
         elif physical_disk.controller == controller.id and \
-                controller.model.startswith("PERC H730P"):
+                controller.model.startswith("PERC H730"):
             return ('/dev/disk/by-path/pci-0000:'
                     '{pci_bus_number}:00.0-scsi-0:'
                     '{channel}:{disk_index}:0').format(
-                pci_bus_number=controller.bus.lower(),
-                channel=2 if physical_disk == "online" else 0,
+                pci_bus_number=pci_bus_number,
+                channel=2 if physical_disk.raid_status == "online" else 0,
                 disk_index=disk_index)
         else:
             return ('/dev/disk/by-path/pci-0000:'
                     '{pci_bus_number}:00.0-sas-0x{sas_address}-lun-0').format(
-                pci_bus_number=controller.bus.lower(),
+                pci_bus_number=pci_bus_number,
                 sas_address=physical_disk.sas_address.lower())
+
+
+def get_pci_bus_number(controller):
+    return controller.bus.lower()
 
 
 def get_fqdd(doc, namespace):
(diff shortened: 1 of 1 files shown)


src/pilot/assign_role.py
```diff
@@ -1094,19 +1101,27 @@ def generate_osd_config_with_journals(controllers, osd_drives, ssds):
...
+                disk_index=disk_index)
+        elif physical_disk.controller == controller.id and \
+                controller.model.startswith("PERC H730P"):
+            return ('/dev/disk/by-path/pci-0000:'
+                    '{pci_bus_number}:00.0-scsi-0:'
+                    '{channel}:{disk_index}:0').format(
+                pci_bus_number=controller.bus.lower(),
+                channel=2 if physical_disk == "online" else 0,
+                disk_index=disk_index)
+        else:
```
cdearborn: The above else will cause the below return to be executed for the first controller in the controller list, which is not what we want.  I think you need to change the above else close to:
elif physical_disk.controller == controller.id:

cdearborn:
I don't think this will work as-is because in the case of single disk RAID0's, it will try to use the physical disks for OSDs instead of the virtual disk RAID0s, but RHEL won't see the physical disks - it will only see the virtual RAID0s.  Have you tested doing a full deployment with this code?  I'm betting that the ceph deployment will fail.
Before going further with this code, I think we should try a full deployment and see if the OSDs are configured correctly.  If not, then we need to investigate the RAID0 virtual disk by-path names in RHEL to determine if we can dynamically create the by-path names from information queried from the iDRAC to make sure generating the OSD configs dynamically is technically feasible.

Provided that it is feasible, I think we're going to need to make the code use for OSDs: any virtual disk that the OS is not installed on, and any physical disks that are in JBOD mode.  As a result, I think we're probably going to have to rewrite select_os_volume() so that it returns the volume that it selected and then pass that volume in to generate_osd_config() so that it knows what volume the OS is installed on so that it can figure out what volumes are available for OSDs.

digambar15:
> Pretty sure you will need to remove the code in generate_osd_config() that causes it to only generate the OSD config if an HBA330 is present. See lines 897-906:
> https://github.com/dsp-jetpack/JetPack/blob/4b7d099ea2ac6156453e5f77c3d7734d3248ac6c/src/pilot/assign_role.py#L897

Yes, I removed it in my testing. Looks like I updated this patch without removing that code.
Will update it

digambar15:
Hey, I have run successful deployment of Overcloud with multiple controllers on Ceph Storage nodes.  I see all the OSD's created successfully with the latest code. Also I have updated code as per our investigation.

src/pilot/assign_role.py
```diff
@@ -274,26 +274,26 @@ def define_target_raid_config(role, drac_client):
 def get_raid_controller_id(drac_client):
     disk_ctrls = drac_client.list_raid_controllers()
 
-    raid_controller_ids = [
-        c.id for c in disk_ctrls if drac_client.is_raid_controller(c.id)]
+    raid_controller_ids = []
 
-    number_raid_controllers = len(raid_controller_ids)
+    for cnt in disk_ctrls:
+        if drac_client.is_raid_controller(cnt.id):
```
cdearborn: I don't think this code will quite work the way you want.  is_raid_controller() will return True for both PERC RAID controllers and BOSS controllers.  See: https://github.com/openstack/python-dracclient/blob/master/dracclient/resources/raid.py#L522
This means that the line above will add the BOSS controller to the end of the list, and the elif will never execute.  If you want the BOSS controllers to be at the beginning of the list, then check to see if the controller is a BOSS first and insert it at the beginning else if it's a RAID controller then add it at the end.  In other words reverses this code so the BOSS check is handled in the "if" and the RAID check is handled in the "elif"

src/pilot/assign_role.py
```diff
@@ -428,9 +459,11 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
     #
     # A successful call returns a list, which may be empty; otherwise,
     # None is returned.
-    jbod_capable = drac_client.is_jbod_capable(raid_controller_name)
+    if not raid_controller:
+        raid_controller = boss_controller
```
cdearborn: This section of code is about defining jbod disks for the remaining disks controlled by the RAID controller that are not being used for the RAID1 to install the OS on.  As it's written, in the BOSS+HBA330 case, it will ask the BOSS if it's JBOD capable below, then call into define_jbod_logical disks below with no remaining_physical_disks.  Instead of doing this, I would rather we just change line 462 above to "if raid_controller:" and delete line 463, then put lines 464-467 inside that "if" statement.  That way, if there is a BOSS+HBA, this code won't be executed, which is what I would expect.

src/pilot/assign_role.py
```diff
@@ -428,9 +459,11 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
+    jbod_capable = drac_client.is_jbod_capable(raid_controller)
     jbod_logical_disks = define_jbod_logical_disks(
-        drac_client, remaining_physical_disks, raid_controller_name,
+        drac_client, remaining_physical_disks, raid_controller,
         jbod_capable)
 
     if jbod_logical_disks is None:
```
cdearborn: This "if" statement doesn't look right to me.  It should be perfectly ok to define an os_logical_disk and no jbod_logical_disks.  This is what would happen in the PERC H740P case, and in the HBA330 case too.  I think lines 469-470 should be deleted.

src/pilot/assign_role.py
cdearborn: We know that the BOSS controller can only control 2 disks, and both of those disks will be used for the OS RAID1, so there is no need to execute this code if a BOSS is present since it will have no other disks to create JBOD RAIDs with.

src/pilot/assign_role.py
```diff
@@ -1094,19 +1120,27 @@ def generate_osd_config_with_journals(controllers, osd_drives, ssds):
...
+                    '{pci_bus_number}:00.0-scsi-0:2:{disk_index}:0').format(
+                pci_bus_number=controller.bus.lower(),
+                disk_index=disk_index)
+        elif physical_disk.controller == controller.id and \
+                controller.model.startswith("PERC H730P"):
+            return ('/dev/disk/by-path/pci-0000:'
+                    '{pci_bus_number}:00.0-scsi-0:'
+                    '{channel}:{disk_index}:0').format(
+                pci_bus_number=controller.bus.lower(),
+                channel=2 if physical_disk == "online" else 0,
```
cdearborn: In the line above, physical_disk is an object, so it will never be equal to "online".  Should this be physical_disk.raid_status?

src/pilot/assign_role.py
```diff
@@ -888,17 +921,6 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index, os_volume_size_gb,
...
-            found_hba = True
-            break
-
-    if not found_hba:
-        LOG.info("Not generating OSD config for {ip} because no HBA330 is "
-                 "present.".format(ip=ip_mac_service_tag))
-        return
-
     LOG.info("Generating OSD config for {ip}".format(ip=ip_mac_service_tag))
     system_id = drac_client.get_system().uuid
```
cdearborn: Does the below work with a PERC H740P RAID controller?  The reason why I ask is that we will create single disk RAID0s in the RAID configuration section, but the get_drives() method below only deals with physical disks and not single disk RAID0s.  I would think get_drives would need to return the single disk RAID0s in this case and not the physical disks that back them.

src/pilot/assign_role.py
```diff
@@ -1094,19 +1120,27 @@ def generate_osd_config_with_journals(controllers, osd_drives, ssds):
...
-
-def get_by_path_device_name(pci_bus_number, physical_disk):
-    return ('/dev/disk/by-path/pci-0000:'
-            '{pci_bus_number}:00.0-sas-0x{sas_address}-lun-0').format(
-                pci_bus_number=pci_bus_number,
+        if physical_disk.controller == controller.id and \
+                controller.model.startswith("PERC H740P"):
+            return ('/dev/disk/by-path/pci-0000:'
+                    '{pci_bus_number}:00.0-scsi-0:2:{disk_index}:0').format(
+                pci_bus_number=controller.bus.lower(),
```
cdearborn: It would be nice to create a get_pci_bus_number() method that takes a single controller and returns controller.bus.lower().  We could then change the 3 locations here to call that method.

src/pilot/assign_role.py
mpardhi23: Yes agreed Chris. I have removed line no. 282 and 283.

src/pilot/assign_role.py
mpardhi23: I have created new method get_pci_bus_number() as suggested.

mpardhi23:
I have done the changes in my local file. Now doing testing of new code changes suggested by you.
Once I done with the testing will push those changes.

src/pilot/assign_role.py
mpardhi23: As discussed in technical call, we have tested the below code with  PERC H740P RAID controller and code is working fine.
We are able to get osd's for  PERC H740P RAID controller with the below code.

src/pilot/assign_role.py
mpardhi23: Agreed Chris. I have updated the code which defining jbod disks for the remaining disks.

mpardhi23:
Hey Chris,
I have addressed suggested comments and tested those changes as well by running overcloud deployment and also by executing assign_role separately on all storage nodes. They are working fine.

src/pilot/assign_role.py
```diff
@@ -274,26 +274,24 @@ def define_target_raid_config(role, drac_client):
 def get_raid_controller_id(drac_client):
```
cdearborn: Would you mind renaming this method to get_raid_controller_ids()?

src/pilot/assign_role.py
```diff
@@ -381,18 +385,39 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
         key=physical_disk_id_to_key)
 
 
-def define_storage_logical_disks(drac_client, raid_controller_name):
+def define_storage_logical_disks(drac_client, raid_controller_ids):
+    boss_controller = None
+    raid_controller = None
```
cdearborn: It would be great if we could make this handle the case where there are multiple RAID controllers.  To do this we could rename this variable to raid_controllers and set it equal to the empty list

src/pilot/assign_role.py
```diff
@@ -381,18 +385,39 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
     all_physical_disks = drac_client.list_physical_disks()
 
+    if len(raid_controller_ids) > 1:
+        for cntrl in raid_controller_ids:
+            if drac_client.is_boss_controller(cntrl):
+                boss_controller = cntrl
+            else:
+                raid_controller = cntrl
```
cdearborn: Here, we could add cntrl to the raid_controllers list

src/pilot/assign_role.py
```diff
@@ -381,18 +385,39 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
+    else:
+        if drac_client.is_boss_controller(raid_controller_ids[0]):
+            boss_controller = raid_controller_ids[0]
+        else:
+            raid_controller = raid_controller_ids[0]
```
cdearborn: Here, we could add raid_controller_ids[0] to the raid_controllers list

src/pilot/assign_role.py
```diff
@@ -381,18 +385,39 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
+
     # Get the drives controlled by the RAID controller
     raid_cntlr_physical_disks = []
```
cdearborn: This would need to change to a dictionary that maps a raid controller id to a list of physical disks.

src/pilot/assign_role.py
```diff
@@ -381,18 +385,39 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
+    boss_cntrl_disks = []
     for disk in all_physical_disks:
-        if disk.controller == raid_controller_name:
+        if disk.controller == boss_controller:
+            boss_cntrl_disks.append(disk)
+        elif disk.controller == raid_controller:
             raid_cntlr_physical_disks.append(disk)
```
cdearborn: Here we would add the disk to the list of disks for the raid controller

src/pilot/assign_role.py
```diff
@@ -381,18 +385,39 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
 
     # Make sure we have enough drives attached to the RAID controller to create
     # a RAID1
-    num_raid_cntlr_physical_disks = len(raid_cntlr_physical_disks)
+
+    if boss_cntrl_disks:
+        num_raid_cntlr_physical_disks = len(boss_cntrl_disks)
+    else:
+        num_raid_cntlr_physical_disks = len(raid_cntlr_physical_disks)
```
cdearborn: as long as one raid controller has at least 2 disks, then we're fine.  if no raid controller has at least 2 disks, then emit the error below

src/pilot/assign_role.py
```diff
@@ -406,8 +431,12 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
         return None
 
-    os_logical_disk = define_storage_operating_system_logical_disk(
-        raid_cntlr_physical_disks, drac_client, raid_controller_name)
+    if boss_controller:
+        os_logical_disk = define_storage_operating_system_logical_disk(
+            boss_cntrl_disks, drac_client, boss_controller)
+    else:
+        os_logical_disk = define_storage_operating_system_logical_disk(
+            raid_cntlr_physical_disks, drac_client, raid_controller)
```
cdearborn: Here, we could just pick the first RAID controller that has 2 or more disks and pass it in including the physical disks that it controls.

src/pilot/assign_role.py
```diff
@@ -428,16 +457,17 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
     #
```
cdearborn: remaining_physical_disks would change to be a clone of the raid_cntlr_physical_disks dict, and the above code would update the list of physical disks for the used RAID controller to not include the disks that we used in the RAID1.

src/pilot/assign_role.py
```diff
@@ -428,16 +457,17 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
     # A successful call returns a list, which may be empty; otherwise,
     # None is returned.
-    jbod_capable = drac_client.is_jbod_capable(raid_controller_name)
-    jbod_logical_disks = define_jbod_logical_disks(
-        drac_client, remaining_physical_disks, raid_controller_name,
-        jbod_capable)
+    logical_disks = [os_logical_disk]
+    if raid_controller:
```
cdearborn: this would change to looping through the raid controllers

src/pilot/assign_role.py
```diff
@@ -1015,9 +1034,17 @@ def get_drives(drac_client):
     spinners = []
     ssds = []
     physical_disks = drac_client.list_physical_disks()
-    for physical_disk in physical_disks:
+    virtual_disks = drac_client.list_virtual_disks()
+    raid_disks = [vd.physical_disks for vd in virtual_disks
+                  if vd.raid_level == '1'][0]
```
cdearborn: If for some reason there are multiple RAID 1s, we don't know which one has the OS on it, so we need to exclude all of the physical disks for all of the RAID1s, otherwise Ceph could overwrite the OS when it initializes the OSDs.  So I think the above needs to change to collect all of the physical disks on all the RAID1s.

src/pilot/assign_role.py
mpardhi23: Done.

src/pilot/assign_role.py
mpardhi23: I have updated code to support multiple controllers.

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Changed the above list of raid_cntrl_physical_disks to dictionary.

src/pilot/assign_role.py
mpardhi23: Done.

src/pilot/assign_role.py
mpardhi23: Agreed. Done the required changes.

mpardhi23:
Hey Chris,
I have addressed all the changes as suggested in review comments.
I have tested the updated code of get_drives() and support of multiple controllers on 13G(R145) and 14G(R160) stamp by running assign_role separately on all storage nodes and also ran overcloud deployment and full deployment as well.
Changes are working fine and ceph installation is successfully done.

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: I have modified the above code as per required changes.

src/pilot/assign_role.py
```diff
@@ -381,22 +385,31 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
-def define_storage_logical_disks(drac_client, raid_controller_name):
+def define_storage_logical_disks(drac_client, raid_controller_ids):
+    raid_controllers = []
     all_physical_disks = drac_client.list_physical_disks()
 
+    if len(raid_controller_ids) > 1:
+        for cntrl in raid_controller_ids:
+            raid_controllers.append(cntrl)
+    else:
+        raid_controllers.append(raid_controller_ids[0])
```
cdearborn: It looks to me like the above code from 392 through 396 is equivalent to a just making a copy of the raid_controller_ids list.  Looking through the code below, it doesn't look like we ever modify the list, so it doesn't appear that there's any need to make a copy.  Could we just eliminate the raid_controllers variable?

src/pilot/assign_role.py
```diff
@@ -406,8 +419,17 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
 
-    os_logical_disk = define_storage_operating_system_logical_disk(
-        raid_cntlr_physical_disks, drac_client, raid_controller_name)
+    if len(raid_controllers) > 1:
+        boss_controller = [cntrl for cntrl in raid_controllers
+                           if drac_client.is_boss_controller(cntrl)][0]
+        if boss_controller:
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[boss_controller],
+                drac_client, boss_controller)
```
cdearborn: let's add an else clause here and pick the first RAID controller in the raid_controllers list that has 2 or more drives, then call into define_storage_operating_system_logical_disk() passing that controller.  This would handle the case of having multiple controllers where neither is a BOSS controller.  I'm honestly not sure that this would be a valid configuration, but I would rather our code do something reasonable in this case just to be sure.  If there are no RAID controllers with 2 or more drives, then we should log the error on line 412 above and return None.

src/pilot/assign_role.py
```diff
@@ -381,22 +385,31 @@ def get_raid_controller_physical_disk_ids(drac_client, raid_controller_fqdd):
...
+            if disk.controller in raid_cntlr_physical_disks:
+                raid_cntlr_physical_disks[disk.controller].append(disk)
+            else:
+                raid_cntlr_physical_disks[disk.controller] = [disk]
 
     # Make sure we have enough drives attached to the RAID controller to create
     # a RAID1
-    num_raid_cntlr_physical_disks = len(raid_cntlr_physical_disks)
-    if num_raid_cntlr_physical_disks < 2:
+    if all(len(raid_cntlr_physical_disks[cntrl]) < 2
```
cdearborn: It's really not necessary for all RAID controllers to have 2 or more disks.  We really just need 1 RAID controller to have 2 or more.  Let's break this code (lines 409 through 413) out into a method that returns True or False and then call that method after we know which controller we're using.  So that would be immediately after line 425 and immediately after line 429.

src/pilot/assign_role.py
```diff
@@ -417,28 +439,29 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
-
-    if jbod_logical_disks is None:
-        return None
-
     logical_disks = [os_logical_disk]
-    logical_disks.extend(jbod_logical_disks)
+    for raid_controller in raid_controllers:
+        jbod_capable = drac_client.is_jbod_capable(raid_controller)
+        jbod_logical_disks = define_jbod_logical_disks(
+            drac_client, remaining_physical_disks, raid_controller,
```
cdearborn: I think there is a bug here.  We check to see if the RAID controller is JBOD capable, then call into define_jbod_logical_disks, but we pass in all of the remaining disks for all RAID controllers.  This means that we may treat some of the disks as being attached to a JBOD capable RAID controller when they are not, or vice versa.  I think we want to pass in the remaining disks for just the RAID controller specified.  Maybe a good way to do that would be to take the raid_cntlr_physical_disks dictionary and remove from it the disks we use for the RAID1.  Then it would have the remaining disks per RAID controller and it could be used here.

src/pilot/assign_role.py
```diff
@@ -1014,34 +1026,58 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
-        # Eliminate physical disks that have an error status
-        if physical_disk.status == 'error':
-            LOG.warning("Not using disk {id}, because it has a status of "
-                        "{status}".format(id=physical_disk.id,
-                                          status=physical_disk.status))
-            continue
+    raid0_disks = [vd for vd in virtual_disks if vd.raid_level != '1']
+
+    raid1_disks = [vd.physical_disks for vd in virtual_disks
+                   if vd.raid_level == '1'][0]
```
cdearborn: I don't think the above 2 lines are right.  It looks like it will make a list of lists of physical drives, and then take the first list of physical drives for the first RAID1.  Since we don't know which RAID1 will be used for the OS drive, we need raid1_disks to be a list of all physical disks for all RAID1s.  I think you want to do something like this:
raid1_disks=[]
for vd in virtual_disks:
  if vd.raid_level == '1':
    raid1_disks.extend(vd.physical_disks)

mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Sure Chris. I have handled the code for multiple raid controllers. And tested those changes with existing controller configuration. It is working fine.

src/pilot/assign_role.py
mpardhi23: Added method to check disk count.

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
```diff
@@ -406,39 +411,50 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
             "configuration")
         return None
 
-    os_logical_disk = define_storage_operating_system_logical_disk(
-        raid_cntlr_physical_disks, drac_client, raid_controller_name)
+    boss_controller = [cntrl for cntrl in raid_controllers
+                       if drac_client.is_boss_controller(cntrl)]
+    if boss_controller and check_cntlr_physical_disks_len(
+         raid_cntlr_physical_disks[boss_controller[0]]):
```
cdearborn: I think there's a potential issue here.  If there is a BOSS present, but it only has 1 drive, then the else clause on line 422 will be entered.  At that point, the first RAID controller that has 2 or more drives will be picked.  I think we should fail in the event that we have a BOSS that has less than 2 drives.  Can you change the above to:
if boss_controller:
    if check_cntlr_physical_disks_len(raid_cntr_physical_disks[boss_controller[0]]):
        ...
    else:
        log a critical error
        return None

src/pilot/assign_role.py
```diff
@@ -406,39 +411,50 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
+            drac_client, boss_controller[0])
+    else:
+        raid_controller = [cntrl for cntrl in raid_controllers
+                           if len(raid_cntlr_physical_disks[cntrl]) >= 2]
+        if raid_controller:
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[raid_controller[0]],
+                drac_client, raid_controller[0])
+        else:
+            LOG.critical("For configuring RAID 1 needs at least two drives")
```
cdearborn: Can you change the above message to:
"At least 2 drives controlled by the same RAID controller are needed to configure a RAID 1"

src/pilot/assign_role.py
```diff
@@ -1014,34 +1019,60 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
+        phy_disk = [pd for pd in physical_disks_of_raid
+                    if virtual_disk.physical_disks[0] == pd.id][0]
+
+        if phy_disk.media_type == 'hdd':
+            spinners.append(virtual_disk)
         else:
-            ssds.append(physical_disk)
+            ssds.append(virtual_disk)
+
+        physical_disks_of_raid.remove(phy_disk)
```
cdearborn: I think there is a bug here.  When we exit this loop, physical_disks_of_raid should contain only physical disks that are not in any RAID for the block of code below to work correctly.  The line above only removes 1 physical disk that backs the current virtual disk from physical_disks_of_raid though, when all of the physical disks that back the current virtual disk should be removed.  I'm thinking that lines 1036 and 1037 should be simplified to just make a list of all physical disks that back the RAID.  I'm thinking that physical_disks_of_raid should be a dictionary that maps pd.id to pd.  That would make lines 1036 & 1037 something like:
phy_disks = [physical_disks_of_raid[pd_id] for pd_id in virtual_disk.physical_disks]
Then we could use phy_disks here to delete all of the disks that back the virtual disk from physical_disks_of_raid.

src/pilot/assign_role.py
```diff
@@ -1014,34 +1019,60 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
+
+    if physical_disks_of_raid:
+        for physical_disk in physical_disks_of_raid:
+            # Eliminate physical disks that in a state other than non-RAID
+            if physical_disk.raid_status != "non-RAID" and \
+               physical_disk.raid_status == "failed":
```
cdearborn: I don't think the above 2 lines are right.  We want to be examining any drive that's in JBOD mode (non-RAID), and any drive connected to an HBA330 (RaidStatus=8, raid_status="non-RAID").  Any drive that's in RAID mode and is not actually in a RAID volume won't be made available to the OS.  So I think we want this to just be:
if physical_disk.raid_status != "non-RAID":
# skip the disk

mpardhi23: We had added physical_disk.raid_status == "failed" condition because earlier there was failed disk on one of the storage node, and in that case get_drives() code was not passing.
To ignore failed disk issues , we have added this condition to skip those disks which are in failed state.

mpardhi23:
I have addressed all your review comments and also tested those changes on 13G and 14G stamp.
Able to get the correct osd configuration and ceph installation is successfully done in full deployment on both of the stamp(R145 and R160).

src/pilot/assign_role.py
```diff
@@ -406,39 +411,56 @@ def define_storage_logical_disks(drac_client, raid_controller_name):
...
+                       if drac_client.is_boss_controller(cntrl)]
+    if boss_controller:
+        if check_cntlr_physical_disks_len(
+           raid_cntlr_physical_disks[boss_controller[0]]):
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[boss_controller[0]],
+                drac_client, boss_controller[0])
+        else:
+            LOG.critical("At least 2 drives controlled by the same RAID "
+                         "controller are needed to configure a RAID 1")
```
cdearborn: Can you change this error message to:
"The BOSS card has only 1 SSD.  2 SSDs are needed to configure a RAID 1"

src/pilot/assign_role.py
```diff
@@ -1014,34 +1025,67 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
...
+            # Eliminate physical disks that in a state other than non-RAID
+            if physical_disks[pd_id].raid_status != "non-RAID":
+                LOG.info("Skipping disk {id}, because it has a RAID status of "
+                         "{raid_status}".format(
+                             id=physical_disks[pd_id].id,
+                             raid_status=physical_disks[pd_id].raid_status))
+                continue
+
+            # Eliminate physical disks that have a failed status
+            if physical_disks[pd_id].raid_status == "failed":
```
cdearborn: This entire "if" clause will never be entered because the "if" clause on line 1053 above will be entered if raid_status=="failed".  Let's just remove all the code from 1060 through 1066 since it won't be used, and modify the comment on line 1052 above to: "Eliminate physical disks in a state other than non-RAID including failed disks"

mpardhi23:
Hey Chris,
I have updated error message and comment. And also removed the code 1060 through 1066.
I would like to test the assign_role with the failed disk scenario as well. Today will ask Calvin to change any disk state to failed and start with the testing. Will update you once I done with this scenario.

mpardhi23:
> Looks good! I'll hold off on merging the PR until your testing is complete. If Calvin doesn't have a failed disk, then you may just want to put in some temporary code that hard sets raid_status to failed on the first virtual disks or something like that for testing purposes.

Thanks Chris! Yes , will test with the temporary code changes for failed disk scenario if Calvin doesn't have failed disk.

mpardhi23:
Hey Chris,
I have tested assign_role with the failed disk scenario by adding some temporary code which hard set the raid_status to 'failed' on one of the drives of storage node. It is skipping that failed disk while generating OSD config and assign_role executed successfully.

PR #406, opened 2019-08-27
CES-10797: DVR Sanity bug fix

zxa13:


gaelrehault:
Not sure this will work looking at config.py, the default value is True (uppercase) when set 
Would work best if in director.py, setup_sanity_ini  you did a lower(str(self.settings.dvr_enable)) ?

zxa13:
> 
> 
> Not sure this will work looking at config.py, the default value is True (uppercase) when set
> Would work best if in director.py, setup_sanity_ini you did a lower(str(self.settings.dvr_enable)) ?

Well at the start of the sanity_test.sh there is DVR_ENABLED=$(get_value_lower dvr_enabled). Here it is already saving the lowercase attribute no matter the case defined in the sanity.ini file.

PR #412, opened 2019-09-03
.fix for satellited locked deployments

gaelrehault:


heydog15:
Looks like you tested it and Rajini validated her Unity container as well.

PR #432, opened 2019-10-10
Kickstart Changes for RHEL8

ahmadfsbd:
Replaced python 2.7 with python 3.6
Firewall service is included by default in rhel8 and disabled in kickstart
List of Packages to be installed is updated accordingly
NTP configurations are replaced by chrony configurations

### DIFF
--- a/src/mgmt/osp-sah.ks
+++ b/src/mgmt/osp-sah.ks
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#version=RHEL7
+#version=RHEL8
 
 install
 cdrom
@@ -67,9 +67,7 @@ eula --agreed
 @virtualization-hypervisor
 @virtualization-tools
 dhcp
-ntp
-ntpdate
--chrony
+chrony
 -firewalld
 system-config-firewall-base
 %end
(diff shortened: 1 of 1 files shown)


src/mgmt/osp-sah.ks
```diff
@@ -366,18 +363,18 @@ do
...
-systemctl enable ntpd
-sed -i -e "/^server /d" /etc/ntp.conf
+# Configure the chrony daemon for ntp
+systemctl enable chronyd
+sed -i -e "/^server /d" /etc/chrony.conf
 
 for ntps in ${NTPServers//,/ }
 do
-  echo "server ${ntps}" >> /etc/ntp.conf
+  echo "server ${ntps}" >> /etc/chrony.conf
```
heydog15: Check to ensure pool isn't the default vs server

src/mgmt/osp-sah.ks
```diff
@@ -596,14 +593,13 @@ fi
 subscription-manager repos ${ProxyInfo} '--disable=*' --enable=rhel-7-server-rpms --enable=rhel-7-server-optional-rpms
 
 systemctl disable NetworkManager
```
gaelrehault: This needs to be removed since we ll now rely on networkManager

src/mgmt/osp-sah.ks
```diff
@@ -596,14 +593,13 @@ fi
...
-systemctl disable firewalld
 systemctl disable chronyd
```
heydog15: Shouldn't this be removed since we need chronyd enabled?

src/mgmt/osp-sah.ks
ahmadfsbd: You are right will change

src/mgmt/osp-sah.ks
ahmadfsbd: Verified, pool and server are both valid. Pool in case of a range of servers and server in case of individual hosts

src/mgmt/osp-sah.ks
ahmadfsbd: Yes , updated.

src/mgmt/osp-sah.ks
```diff
@@ -595,15 +591,11 @@ fi
 
 subscription-manager repos ${ProxyInfo} '--disable=*' --enable=rhel-7-server-rpms --enable=rhel-7-server-optional-rpms
```
heydog15: Change the repos to --enable=rhel-8-for-x86_64-baseos-rpms --enable=rhel-8-for-x86_64-appstream-rpms

heydog15:
Looks good.

PR #434, opened 2019-10-16
Included SAH-Network Configurations through Kickstart's Network Commands

ahmadfsbd:
Network configurations are now performed through Kickstart's "network" commands
Configurations have been moved to %pre section where they get included in ks_include.txt
Updated chrony configurations.

In %post section, a work around is included for following issues:
Vlans not being enslaved by bridges in %pre section
STP=yes
ONBOOT=no for prov and pub-api vlans

### DIFF
--- a/src/deploy/osp_deployer/settings/sample.properties
+++ b/src/deploy/osp_deployer/settings/sample.properties
@@ -6,14 +6,14 @@
         "root_password": "xxxxxxxxxx",
 
         "anaconda_ip":"192.168.190.134",
-        "anaconda_iface":"em2.190",
+        "anaconda_iface":"eno4",
 
         "public_bond": "bond1",
-        "public_slaves": "em2 p1p2",
+        "public_slaves": "eno2,enp4s0f1",
         "public_api_ip":"192.168.190.12",
 
         "private_bond": "bond0",
-        "private_slaves": "em1 p1p1",
+        "private_slaves": "eno1,enp4s0f0",
         "provisioning_ip":"192.168.120.12",
         "storage_ip":"192.168.170.12",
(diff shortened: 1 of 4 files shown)


src/deploy/osp_deployer/settings/sample.properties
```diff
@@ -9,11 +9,11 @@
         "anaconda_iface":"em2.190",
 
         "public_bond": "bond1",
-        "public_slaves": "em2 p1p2",
+        "public_slaves": "em2,p1p2",
```
heydog15: Are these the correct interface names for RHEL8?  I thought they were the ensXXX type of names.

src/mgmt/osp-sah.ks
```diff
@@ -119,41 +118,41 @@ anaconda_interface="CHANGEME e.g. 10.148.44.211/255.255.255.0 em4 no"
 # Define the bonds
 # Bond to handle external traffic
 extern_bond_name="CHANGEME e.g. bond1"
-extern_boot_opts="onboot none"
+extern_boot_opts="onboot"
 extern_bond_opts="mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1"
-extern_ifaces="CHANGEME e.g. em2 p1p2"
+extern_ifaces="CHANGEME e.g. em2,p1p2"
```
heydog15: Same question as the comment in the sample file for the names.

src/deploy/osp_deployer/settings/sample.properties
ahmadfsbd: Updated the SAH interface names for RHEL8 in sample files. Thank you for pointing it out

PR #448, opened 2019-11-15
T10 PI format patch

zxa13:
T10 PI format of drives are causing failures deploying solutions. For the fix I added this  'mpt3sas.prot_mask=1' parameter in the 'src/pilot/templates/first-boot.yaml' in the CEPHStorage section so that this parameter is only added to the storage nodes and not applied to all the nodes in the architecture. To verify post-deployment follow these steps:

- ssh in storage node
- goto file /etc/default/grub
- verify if GRUB_CMDLINE_LINUX = 'mpt3sas.prot_mask=1 ...' is set
- also verify /boot/efi/EFI/redhat/grub.cfg

### DIFF
--- a/src/pilot/templates/first-boot.yaml
+++ b/src/pilot/templates/first-boot.yaml
@@ -178,7 +178,7 @@ resources:
             echo $cmd
             $cmd
             echo
-
+            
           done
           partprobe
           parted -lm
--- a/src/pilot/customize_image.sh
+++ b/src/pilot/customize_image.sh
@@ -144,6 +144,10 @@ if [ ! -z "${satellite_hostname}" ]; then
 else
    run_command "virt-customize -a overcloud-full.qcow2 --run-command \"echo '${director_ip} ${director_short} ${director_long}' >> /etc/hosts\""
 
+   #Patch for configuring T10 PI format drives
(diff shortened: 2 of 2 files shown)


zxa13:
Hey Chris. So I the command is showing me all "none"s but I cant find the changes in the grub cfg file. After redeploying my

src/pilot/templates/first-boot.yaml
```diff
@@ -178,7 +178,7 @@ resources:
             echo $cmd
             $cmd
             echo
-
+
```
cdearborn: The extra spaces may be causing an issue...

zxa13: I checked with the previous files and even in this file that we are using spaces to make the code more readable but yes I will look into this.

cdearborn: I think having extra blank lines is ok, but it may be that having any whitespace in a blank line is not.

zxa13: The ansible playbook issue was with the original first-boot.yaml file in which I added the kernel parameters. I checked and there doesnt seem to any extra whitespaces. I am gonna keep looking into it. Have restarted the deployment on R8. Lets see how that goes.

PR #449, opened 2019-11-19
Changes for creating rhel 8 based Director VM 

ahmadfsbd:
This involves the following
1- Changes in director kickstart (packages, interface names, chrony, vm specs etc.)
2- Python3 shebang changes
3- Changes for decoding default python3 encoded strings
4- Updated rpms for rhel8 and osp15 based deployment
5-  Updated python path in sah kickstart

### DIFF
--- a/src/deploy/auto_common/ipmi.py
+++ b/src/deploy/auto_common/ipmi.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright (c) 2015-2019 Dell Inc. or its subsidiaries.
 #
@@ -24,26 +24,6 @@ logger.setLevel(logging.WARNING)
 
 
 class Ipmi:
-    """
-    this assumes CygWin is installed along with ipmi
-    (if used on windows)
-    ipmitools : download & extract from
-        http://sourceforge.net/projects/ipmitool/files/
-        ./configure --enable-intf-lanplus
-        check the output contains :
-        Interfaces
-        lan : yes
-        lanplus : yes
-        if not, check the log output,
-            and fix the missing dependencies etc.
-        make
-        make install
-        Add c:\\cygwin\\bin and c:\\cygwin\\usr\\local\\bin to your path
-        (or the relevant paths for your setup)
-        make sure you can talk to one of your servers :
-        in cygwin :
-        ipmitool -I lanplus -H idracIp -U user -P password  power status
-    """
 
     def __init__(self, cygwinlocation, ipmi_user, ipmi_password, idrac_ip):
         self.cygwin_loc = cygwinlocation
--- a/src/deploy/osp_deployer/checkpoints.py
+++ b/src/deploy/osp_deployer/checkpoints.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2015-2019 Dell Inc. or its subsidiaries.
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
+# Licensed under the Apache License, Version 2.0 (the "License");# you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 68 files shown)


src/deploy/osp_deployer/settings_sanity.py
```diff
@@ -27,7 +27,7 @@
 
 class DeployerSanity:
     def __init__(self):
-        self.settings = Settings.settings
+        self.settings = Settings("/root/r7_custom.ini")
```
heydog15: Why do we have an internal rack name here?

src/mgmt/deploy-director-vm.sh
```diff
@@ -336,8 +322,8 @@ rm -f /store/data/images/director.img
     --vcpus 8 \
     --hvm \
     --os-type linux \
-    --os-variant rhel7 \
-    --disk /store/data/images/director.img,bus=virtio,size=80 \
+    --os-variant rhel8.1 \
```
heydog15: Do we have to spell out the sub version, e.g, rhel8.1 vs rhel8?  Seems like we should stick with just rhel8.

src/deploy/osp_deployer/settings_sanity.py
ahmadfsbd: that was left by mistake. will correct it. Thanks

src/mgmt/deploy-director-vm.sh
```diff
@@ -357,8 +343,8 @@ virt-install --name director \
   --vcpus 8 \
   --hvm \
   --os-type linux \
-  --os-variant rhel7 \
-  --disk /store/data/images/director.img,bus=virtio,size=80 \
+  --os-variant rhel8.1 \
```
heydog15: Do we have to spell out the sub version, e.g, rhel8.1 vs rhel8?  Seems like we should stick with just rhel8.

src/mgmt/deploy-director-vm.sh
ahmadfsbd: we don't have to be specific. will just change it to 8 then

src/deploy/auto_common/ipmi.py
```diff
@@ -38,7 +39,7 @@ class Ipmi:
             and fix the missing dependencies etc.
         make
         make install
-        Add c:\\cygwin\\bin and c:\\cygwin\\usr\\local\\bin to your path
+        Add r"c:\\cygwin\\bin" and r"c:\\cygwin\\usr\\local\\bin" to your path
         (or the relevant paths for your setup)
         make sure you can talk to one of your servers :
         in cygwin :
```
gaelrehault: We can actually remove all of the above, this is back from the (very old) days we were supporting installations from windows.

src/deploy/osp_deployer/checkpoints.py
```diff
@@ -160,7 +158,7 @@ def sah_health_check(self):
                 Ssh.execute_command(self.sah_ip,
                                     "root",
                                     self.settings.sah_node.root_password,
-                                    cmd)[1]:
+                                    cmd)[1].decode('utf-8'):
```
gaelrehault: Wouldn t it be more convenient to decode('utf-8') the string in the Sssh.execute_command rather than everytime we call it ?

ahmadfsbd: Will be updating the code for that. Thank you

src/deploy/auto_common/ipmi.py
gaelrehault: I meant the entire comment section can be removed, ipmi ships with rhel the install instructions are no longer relevant.

ahmadfsbd: The comment section has been removed from the code.

src/deploy/osp_deployer/checkpoints.py
ahmadfsbd: Updated

src/deploy/osp_deployer/settings_sanity.py
ahmadfsbd: Updated

src/mgmt/deploy-director-vm.sh
ahmadfsbd: Just verified that in rhel8 you have to be specific with the os name and it does not accept just any name. Used "osinfo-query os | grep rhel" command to list down available os names

PR #460, opened 2019-12-30
Disabled realtime drives conversion from jetpack and removed drives conversion in raid from Ironic

Rachit7194:
Hi Chris,

In this PR, i have removed drives conversion(in raid before creating raid config) from ironic in (raid.patch file) as we already discussed that we should not remove drives conversion from Jetpack and osp release should itself put that drives conversion.

Also i have disable again drives conversion in realtime mode as some servers does not let drives convert in either mode in realtime.

Thanks,
-Rachit


Rachit7194:
Sure Chris, 
Thanks for the review

PR #467, opened 2020-01-10
CES-10988 

gaelrehault:
adding ironic python dependencies to undercloud host
python3 edits

### DIFF
--- a/src/pilot/config_idrac.py
+++ b/src/pilot/config_idrac.py
@@ -657,7 +657,7 @@ def config_idrac(instack_lock,
                     new_drac_client.is_idrac_ready()
                     password_changed = True
                 except exceptions.WSManInvalidResponse as ex:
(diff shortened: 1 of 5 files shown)


src/pilot/config_idrac.py
```diff
@@ -657,7 +657,7 @@ def config_idrac(instack_lock,
                     new_drac_client.is_idrac_ready()
                     password_changed = True
                 except exceptions.WSManInvalidResponse as ex:
-                    if "unauthorized" in ex.message.lower():
+                    if "unauthorized" in str(ex).lower():
```
heydog15: Why the change here?  Is WSMan changing?

gaelrehault: Python3 handling or exceptions changed

heydog15:
Ok, thanks.

PR #471, opened 2020-01-17
Added : RaidStatus with RAIDStatus attribute

Rachit7194:
Hi Chris, 
I have added a change for which if RAIDStatus is not found then it will take RaidStatus.

Please have a look.

Thanks, 
Rachit

### DIFF
--- a/src/pilot/dracclient_raid.patch
+++ b/src/pilot/dracclient_raid.patch
@@ -1,5 +1,5 @@
---- /usr/lib/python2.7/site-packages/dracclient/resources/raid.py	2019-06-04 04:56:04.000000000 +0000
-+++ raid.py	2019-11-20 18:29:37.056445258 +0000
+--- raid.py.orig	2019-06-04 04:56:04.000000000 +0000
++++ raid_new.py	2020-01-17 10:55:10.617055104 +0000
 @@ -12,6 +12,7 @@
  #    under the License.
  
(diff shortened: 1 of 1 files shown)


src/pilot/dracclient_raid.patch
```diff
@@ -18,7 +18,16 @@
  class PhysicalDisk(PhysicalDiskTuple):
  
      def __new__(cls, **kwargs):
-@@ -633,15 +637,17 @@
+@@ -232,6 +236,8 @@
+         size_b = self._get_virtual_disk_attr(drac_disk, 'SizeInBytes')
+         drac_status = self._get_virtual_disk_attr(drac_disk, 'PrimaryStatus')
+         drac_raid_status = self._get_virtual_disk_attr(drac_disk, 'RAIDStatus')
++        if not drac_raid_status:
```
cdearborn: Hey Rachit,

I don't think this will work because the code will throw an exception if the RAIDStatus attribute is not present.  See:
https://github.com/openstack/python-dracclient/blob/master/dracclient/resources/raid.py#L204
and
https://github.com/openstack/python-dracclient/blob/091b5167258f521deba3594f93fc199182612c22/dracclient/utils.py#L95

Rachit7194: Hi Chris, 
I got your point. I did not test this changes. Actually i forget to mention that "i have not tested this changes" in description and also i was not aware of the exception. Apologies..!.
Currently i am blocked on issue of node getting stuck in clean wait while testing(experimental_upstream_ironic and master) merge changes. So i was not able to reach at that point where this method gets invoke.
But after resolving above changes, i will make sure to test this PR changes and will you know here."

cdearborn: Thanks Rachit!

Rachit7194:
Hi Chris,

I have added another change which handled missing attribute which raise an error, as you suggested.

Currently i have not tested against deployment, But i found that current change is in  _parse_drac_virtual_disk(self, drac_disk)  at line - [L173](https://github.com/openstack/python-dracclient/blob/master/dracclient/resources/raid.py#L173) and this method is called in list_virtual_disks() at line [L170](https://github.com/openstack/python-dracclient/blob/master/dracclient/resources/raid.py#L170).
So i called list_virtual_disk() method by importing dracclient normal python script and tested this change. It works successfully and gives None without raising exception if attribute is invalid.

However, I just resolved the issue of nodes getting stuck in clean_wait after merging experimental_upstream_ironic and master branch. By tomorrow i will test this change by running assign_role or an overcloud deployment. Please do not merge.

Thanks and Regards,
Rachit

cdearborn:
This looks much better.  Thanks!

Rachit7194:
Hi Chris,

I have updated ``RAIDStatus`` attribute to use ``RaidStatus`` attribute if not found.
Also i have tested this change on single node by upgrading its LCC firmware version to latest(4.00.00) 
and it use ``RAIDStatus``  attribute and executes successfully.
I have testing on single node by running assign_role.py and config_idrac.py
Let me know if you want me to do an overcloud deployment.

Thanks and Regards,
Rachit

PR #475, opened 2020-01-22
Update: use RaidStatus if RAIDStatus node found

cdearborn:
This is a temporary patch to make JetStream work with LCC 4.00.00.00


heydog15:
Looks good.

PR #478, opened 2020-01-28
Increase timeout to wait for iDRAC reboot to finish after reset

cdearborn:
This patch increases the time we wait before checking to see if the iDRAC is ready after resetting the iDRAC and it becomes pingable.


heydog15:
Looks good.

PR #481, opened 2020-01-29
Bluestore revert

gaelrehault:
Removing workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1666822 that has been fixed upstream


heydog15:
Tested this change successfully on R8.  Looks good.

PR #482, opened 2020-01-29
moved drives conversion from jetpack to ironic

Rachit7194:
Hi Chris,

As we discuss over call, i have added drives conversion monkey patch for the two upstream ironic patches below.

- DRAC: Drives conversion from raid to jbod: 
          - https://review.opendev.org/#/c/671038/

- DRAC: Drives conversion from JBOD to RAID
         -  https://review.opendev.org/#/c/679093/

I will remove this monkey patch once osp add this changes to ironic.

Please have a look.

Thanks and Regards,
Rachit

### DIFF
--- a/src/pilot/raid.patch
+++ b/src/pilot/raid.patch
@@ -1,14 +1,15 @@
---- raid_original.py	2019-12-30 10:29:42.226859376 +0000
-+++ raid_modified.py	2019-12-30 10:29:18.381080327 +0000
-@@ -15,6 +15,7 @@
- DRAC RAID specific methods
- """
+--- raid_orignal.py	2020-01-27 14:20:46.457601298 +0000
++++ raid_new.py	2020-01-27 14:20:01.905979038 +0000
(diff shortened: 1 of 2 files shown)


src/pilot/raid.patch
```diff
@@ -51,7 +52,23 @@
  def create_virtual_disk(node, raid_controller, physical_disks, raid_level,
                          size_mb, disk_name=None, span_length=None,
                          span_depth=None):
-@@ -202,7 +232,108 @@
+@@ -159,7 +190,14 @@
+     drac_job.validate_job_queue(node)
+ 
+     client = drac_common.get_drac_client(node)
+-
++    LOG.info("Properties before creating VD")
```
cdearborn: This looks like debugging code.  This code isn't present upstream.

cdearborn: This patch also includes the changes for the drive size changing during conversion.
I'll go ahead and merge this, but I think this patch should be revisited when the backports are complete.

PR #486, opened 2020-01-31
CES-11040 - CES-11019

gaelrehault:
- removing dashboard
- removed dns addition to subnet (already created by undercloud install)
- added quotes to images yaml settings


src/pilot/install-director.sh
```diff
@@ -254,12 +254,6 @@ done
 create_flavor baremetal false
 echo "## Done."
```
heydog15: Does this code snipet pertain to deploying Dashboard?

gaelrehault: not to do with dashboard

PR #488, opened 2020-02-06
Nights watch model PXE NIC update

zxa13:
The pxe_nic slot has changed since we acquired the OCP cards for the XE2420 (Nights Watch).


heydog15:
Looks good.  That's what I tested with.

PR #493, opened 2020-02-13
NVMe support in generation of Ceph-OSD configuration

spranjali:
Hey @cdearborn , 
This PR is having implementation of auto-generation of Ceph OSD configuration for all including NVMe drives where execution of assign_role.py ensures that the required NVMe drives are also taken into configuration yaml file for Ceph installation instead of explicit mention in .properties file.
Tested this code on below configurations -
1. Storage node with HDDs and SSDs
2. Storage node with SSDs and NVMes

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -959,10 +959,7 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
 def get_drives(drac_client):
     spinners = []
     ssds = []
-    # Getting all NVMe drives
-    nvme_drives = [drives for drives in drac_client.list_physical_disks()
-            if drives.device_protocol is not None  and
-            drives.device_protocol.startswith("NVMe")]
+    nvme_drives = []
 
     virtual_disks = drac_client.list_virtual_disks()
 
@@ -988,6 +985,9 @@ def get_drives(drac_client):
 
     if physical_disks:
         for pd_id in physical_disks:
+            # Get all NVMe drives
+            if is_nvme_drive(physical_disks[pd_id]):
+                nvme_drives.append(physical_disks[pd_id])
             # Eliminate physical disks in a state other than non-RAID
             # including failed disks
             if physical_disks[pd_id].raid_status != "non-RAID":
@@ -1022,14 +1022,13 @@ def get_drives(drac_client):
     return spinners, ssds, nvme_drives
 
 def generate_osd_config_without_journals(controllers, drives):
-
     osd_config = {
         'osd_scenario': 'lvm',
         'osd_objectstore': 'bluestore',
         'devices': []}
     for drive in drives:
         # Get by-path device name for NVMe drives
-        if drive.controller.startswith('PCIe'):
+        if is_nvme_drive(drive):
             nvme_device_name = get_by_path_nvme_device_name(drive)
             osd_config['devices'].append(nvme_device_name)
         else:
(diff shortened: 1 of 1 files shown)


src/pilot/assign_role.py
```diff
@@ -959,6 +959,11 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
 def get_drives(drac_client):
     spinners = []
     ssds = []
+    # Getting all NVMe drives
+    nvme_drives = [drives for drives in drac_client.list_physical_disks()
+            if drives.device_protocol is not None  and
+            drives.device_protocol.startswith("NVMe")]
```
cdearborn: Hey Pranjali, we always try to limit the number of calls we make to the iDRAC as much as possible.  There is a call to list_physical_disks() on line 963 above and also one on line 976 below when I think we could retrieve the physical disks just one time.  I think I would move this code into the "for" loop below, just below this line "for pd_id in physical_disks:".
You could check the protocol like you do above and if the drive is an NVMe drive, add it to an "nvmes" list, just like we do with spinners and ssds.

src/pilot/assign_role.py
```diff
@@ -1023,9 +1028,14 @@ def generate_osd_config_without_journals(controllers, drives):
         'osd_objectstore': 'bluestore',
         'devices': []}
     for drive in drives:
-        drive_device_name = get_by_path_device_name(
-             drive, controllers)
-        osd_config['devices'].append(drive_device_name)
+        # Get by-path device name for NVMe drives
+        if drive.controller.startswith('PCIe'):
```
cdearborn: I think we should always use the same method to determine if a drive is an NVMe drive in all cases.  Here's we check the controller, but above, we check the device_protocol.  I think the better way to do it is to check the device_protocol.  Can you make a new method called is_nvme_drive() and perform the check in there?  Then you can call it from both places, and we can be sure that NVMe drives are detected the same way everywhere.

src/pilot/assign_role.py
spranjali: Sure Chris. It will be incorporated as recommended.

src/pilot/assign_role.py
spranjali: Hey Chris, 
Check for NVMe drives from get_drives() will be done through device_protocol as it is done for only physical disks while checking NVMe drives from generate_osd_config_without_journals() cannot be done on similar grounds as the drives would be sum of physical disks and virtual disks where virtual disks have no attribute named "device_protocol". Therefore in this particular function drive.controller is checked instead of drive.device_protocol.

However , I tried implementing your review comment and tested positively but not sure whether it will be all-time valid condition. It is -

def is_nvme_drive(disk):
    return True if (not disk.id.startswith("Disk.Virtual") and disk.device_protocol and disk.device_protocol.startswith("NVMe") else False

So when is_nvme_drive() is called from generate_osd_config_without_journals() , it will only extract PCIeSSD disks from a list of physical disks and virtual disks.

src/pilot/assign_role.py
```diff
@@ -983,6 +985,9 @@ def get_drives(drac_client):
 
     if physical_disks:
         for pd_id in physical_disks:
+            # Get all NVMe drives
+            if is_nvme_drive(physical_disks[pd_id]):
+                nvme_drives.append(physical_disks[pd_id])
```
cdearborn: I think you need a continue above.  Otherwise, line 993 will be executed below.  The raid_status will be None for NVMe drives, so the if statement will be entered, then it will log the skipping message even though it did not skip the drive, then continue.

src/pilot/assign_role.py
```diff
@@ -1056,6 +1065,14 @@ def generate_osd_config_with_journals(controllers, osd_drives, ssds):
 
     return osd_config
 
+def is_nvme_drive(disk):
```
cdearborn: This looks good.  Can you add a comment here that says something like:
# This method could be called with either physical or virtual disks, and only physical disks have the device_protocol attribute.  Virtual disks can never be NVMe drives, so only return True if device_protocol is present and indicates NVMe.

src/pilot/assign_role.py
spranjali: Done.

src/pilot/assign_role.py
spranjali: Done.

PR #498, opened 2020-02-21
updated ceph-ansible repo

gaelrehault:


heydog15:
Why does config.py have Openstack 13? Is that ok for  OSP 16 JetPack?

heydog15:
Looks good.

PR #499, opened 2020-02-27
[CES-11041] Sanity testing for SRIOV Offload enabled VM

zxa13:


src/pilot/deployment-validation/sanity_test.sh
```diff
@@ -1119,4 +1122,4 @@ else
   info "##### Done #####"
 
   exit
-fi
+fi
```
dpaterson: Please end files with LF

dpaterson: Please end files with LF

dpaterson: Please end files with LF

PR #501, opened 2020-02-27
Add CLI option to skip iDRAC configuration

dpaterson:
If you are doing lots of re-deployments there is no need to always configure the iDRACs.  This argument let's you turn off that feature, saving time.

### DIFF
--- a/src/deploy/osp_deployer/deployer.py
+++ b/src/deploy/osp_deployer/deployer.py
@@ -55,6 +55,10 @@ def get_settings():
                         help='Do not reinstall the Dashboard VM',
                         action='store_true',
                         required=False)
+    parser.add_argument('-skip_idrac_config', '--skip_idrac_config',
+                        help='Skip configuring iDRACs',
+                        action='store_true',
+                        required=False)
     group = parser.add_mutually_exclusive_group()
     group.add_argument('-validate_only', '--validate_only',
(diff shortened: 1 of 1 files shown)


gaelrehault:
Could we do this in the .properties on a per node basis instead ? similar to existing skip_raid_config, skip_raid_config, skip_nic_config, pxe_nic optional properties

heydog15:
Agree with Gael's coments.

src/deploy/osp_deployer/deployer.py
```diff
@@ -215,7 +221,8 @@ def deploy():
         director_vm.setup_net_envt()
         director_vm.configure_dhcp_server()
         director_vm.node_discovery()
-        director_vm.configure_idracs()
+        if args.skip_idrac_config is False:
+            director_vm.configure_idracs()
```
cdearborn: I think the way this is implemented will be a bit confusing for users, since if skip_idrac_config is True, then it will skip NIC configuration too regardless of what skip_nic_config is set to.  I think it would be better to only skip config_idracs.py if skip_idrac_config is True so that the 2 flags work independently.

dpaterson: Makes sense, will do.

dpaterson:
Closing for now, will revisit this change in an future PR

PR #502, opened 2020-02-27
Support For powerMax

rajinir:


src/pilot/deploy-overcloud.py
```diff
@@ -579,6 +586,14 @@ def main():
         if args.enable_unity_manila:
             env_opts += " -e ~/pilot/templates/unity-manila-config.yaml"
 
+        if args.enable_powermax:
+           if self.settings.powermax_storage_protocol == "iSCSI":
```
gaelrehault: Deploy-overcloud.py will not resolve self.settings.powermax_storage_protocol, you probably want to pass which protocol to use as another param instead

rajinir: Fixed it

rajinir:
Tested this on R11 stamp, deployment was successful and sanity test passes

gaelrehault:
Rajini, could you update the sample files Storage Backends comments section to reflect the options ( and the 2 protocoles)

src/deploy/osp_deployer/settings/config.py
```diff
@@ -418,6 +418,28 @@ def __init__(self, settings_file):
...
+        if backend_settings['enable_powermax_backend'].lower() == 'true':
+            self.enable_powermax_backend = True
+            self.powermax_backend_name = backend_settings['powermax_backend_name']
+            self.powermax_san_ip = backend_settings['powermax_san_ip']
+            self.powermax_san_login = backend_settings[
+                'powermax_san_login']
+            self.powermax_san_password = backend_settings[
+                'powermax_san_password']
+            self.powermax_protocol = backend_settings[
+                'powermax_protocol']
```
gaelrehault: Probably want to check that protocol being set is a valid one. ( especially with mixed case in iSCI)

rajinir:
> Rajini, could you update the sample files Storage Backends comments section to reflect the options ( and the 2 protocoles)

Done.

PR #505, opened 2020-03-05
CES-11082 adding rhosp release info to logs

gaelrehault:


heydog15:
Do we somehow want to state it is also the overcloud version, not just the undercloud?  Maybe add it to the useful IP's section we spit out?

gaelrehault:
_Maybe add it to the useful IP's section we spit out?_  that's done in this PR.

heydog15:
Looks good.

PR #510, opened 2020-03-11
YAQL expression parsing error fix

zxa13:
Empty DellComputeParameters section was causing YAQL expression error during overcloud deployment. Now this DellComputeParameters section will only be used when we are setting KernelArgs parameter.


heydog15:
+2

PR #531, opened 2020-05-12
updated cinder-volume container #

gaelrehault:


heydog15:
Looks good.

rajinir:
How about manila? Did we update that as well?

PR #539, opened 2020-06-08
Patch director utf8 issue

dpaterson:


heydog15:
Looks good.  That's what we tested with.

PR #543, opened 2020-06-10
Monkey patch for ehba support in ironic

Rachit7194:
Hello Chris,
This change is regarding the ironic upstream patch[730725](https://review.opendev.org/#/c/730725) which add support for RAID controller to convert  to RAID mode from EHBA mode.

Ironic patch is currently not merged. But changes are approved.

Thanks,
Rachit


Rachit7194:
> Hey Rachit, this patch doesn't look quite right. The upstream patch uses "is_ehba_mode" while this one uses "is_ehab". The upstream patch defines and uses _EHBA_MODE while this one has that hardcoded. It doesn't look like this patch was generated using the latest upstream code.

Thanks addressed suggested change.

PR #557, opened 2020-07-06
Sample ini was missing some of the edge subnet attributes

dpaterson:


src/deploy/osp_deployer/settings/sample_xsp_profile.ini
```diff
@@ -652,14 +652,72 @@ node_types=
...
 # inspection_iprange = 192.168.122.51,192.168.122.120
 # gateway = 192.168.122.1
+# nic_port_count = 5
+# When using any 4 port NIC configuration, comment out or delete the
+# ProvisioningInterface line below.
+# ProvisioningInterface=em1
+# Bond0Interface1=p1p1
+# Bond0Interface2=p3p1
+# Bond1Interface1=p1p2
+# Bond1Interface2=p3p2
```
gaelrehault, line 703: we should probably use new nic format in samples ?

heydog15: Not for JS-13.3 branch.  Agree for JS-16.1 though

dpaterson: 16.1 will have correct interfaces

PR #561, opened 2020-07-07
PowerFlex  Settings support

rajinir:
The first draft of ini file settings for VxFlexOS. Have deployed and tested it.

### DIFF
--- a/src/deploy/osp_deployer/settings/sample_csp_profile.ini
+++ b/src/deploy/osp_deployer/settings/sample_csp_profile.ini
@@ -620,3 +620,48 @@ node_types=
 # dhcp_end = 192.168.122.250
 # inspection_iprange = 192.168.122.51,192.168.122.120
 # gateway = 192.168.122.1
+
+[PowerFlex Settings]
+
+# Settings for PowerFlex OS Storage Backend
+enable_powerflex_backend=false
+
+# Method to retrieve RPMs before Installation begins
+# Can be either local or external
+powerflex_rpms_method=local
+
+# Used to override default RPMs path.
+# PowerFlexRpmsPath='/opt/dellemc/powerflex/rpms'
+
+# Name of the PowerFlex OS cluster
+powerflex_cluster_name=powerflex
+
+# Name of the  PowerFlex OS Protection Domain
+# A single protection is defined for the whole system.
+powerflex_protection_domain=domain1
+
+# Name of the  PowerFlex OS Storage Pool
+# A single storage pool is defined where all disks of the same type and speed will reside.
+powerflex_storage_pool=pool1
+
+# PowerFlex OS Cluster Configuration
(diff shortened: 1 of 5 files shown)


src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -620,3 +620,56 @@ node_types=
...
+
+# Used to override default RPMs path.
+# PowerFlexRpmsPath = '/opt/dellemc/powerflex/rpms'
+
+# Name of the PowerFlex OS cluster
+powerflex_cluster_name = powerflex
+
+# Name of the  PowerFlex OS Protection Domain
+# A single protection is defined for the whole system.
+powerflex_protection_domain = domain1
```
gaelrehault, line 641: Dont think there should be spaces in all of those settings , regardless of them being picked up within the String values (which would cause issues), they should be formatted the same way as the rest of the settings

rajinir: > Dont think there should be spaces in all of those settings , regardless of them being picked up within the String values (which would cause issues), they should be formatted the same way as the rest of the settings

Done

src/deploy/osp_deployer/settings/sample_csp_profile.ini
```diff
@@ -620,3 +620,56 @@ node_types=
...
+powerflex_cluster_virtualip=192.168.170.40
+
+# By default, All traffic including SDS-SDS and SDS-SDC goes to the same network
+# Uncomment the below line when using dedicated network rebuild and rebalance which can
+# traffic intensive based upon the eventual failure of a SDS node. (Optional)
+# powerflex_rebuildinterface
+
+# List of comma separated disks which will be used as the backend for SDSs.
+powerflex_disks=['/dev/nvme0n1','/dev/nvme1n1','/dev/nvme2n1','/dev/nvme3n1','/dev/nvme4n1','/dev/nvme5n1','/dev/nvme6n1','/dev/nvme7n1']
+
```
gaelrehault, line 669: Where are those disks located ?

jproque-dell: Those disks are hosted on the storage nodes or wherever the SDS service is installed. Think of HCI model where SDS and SDC coexist.

rajinir: Storage Nodes... Mixed mode is not supported yet. The playbooks assume the disk set is same for all three storage nodes

jproque-dell: As long as can get a comma separated list of values,.

rajinir: > As long as can get a comma-separated list of values,.

Will remove this setting then

PR #566, opened 2020-07-16
Fixing comments

rajinir:


heydog15:
Do we want to keep the Manila for Powermax in there as well?  Didn't think we supported that yet either.

rajinir:
> Do we want to keep the Manila for Powermax in there as well? Didn't think we supported that yet either.

Removed it. See updated patch

heydog15:
Looks good.

PR #573, opened 2020-08-04
updated samples 's repos

gaelrehault:


heydog15:
So where do we use the Ceph4 repos?  Is that all contained in the overcloud image?

PR #577, opened 2020-08-07
CES-11250 part II

gaelrehault:
added patches to fix an issue with nova scheduler preventing instances creation (bz 1846020)


heydog15:
Looks good.  Tested on Mobile Rack.

PR #580, opened 2020-08-11
Updated NFV nova scheduler filters

ahmadfsbd:
- Some outdated filters (e.g RamFilter) are removed

- Some additional filters added for ovs-hw-offload


heydog15:
Merging per Sambhu's review.

PR #591, opened 2020-08-31
Updated .ini file for rhel82 and updated Satellite doc path

heydog15:


gaelrehault:
@heydog15 could you also update the sample_hci_profile.ini

heydog15:
Update the sample_hci_profile.ini as well.

PR #593, opened 2020-09-03
Don't use FORCE when clearing the job queue

cdearborn:
Clearing the job queue with FORCE causes a service internal to the iDRAC
to restart.  Until this service is fully up, information queried from
the iDRAC may not be complete. This patch changes config_idrac to
not use the FORCE option to resolve related issues.


heydog15 (approved):

heydog15:
Tested this PR by running 4 installs with no issues on the Mobile Rack.

PR #597, opened 2020-09-07
DVR parameter override fix 

ahmadfsbd:
Forgot to include the re-ordering of DVR template in previous DVR pull request. This will override neutron-ovs.yaml parameters to deploy DVR.


ahmadfsbd:
Will send another one with just one change. All the changes were already merged

PR #599, opened 2020-09-07
SRIOV-Offload support, updates and templates

ahmadfsbd:
- Support for sriov ovs-hw-offload in JS16.1
- Removes the additional dedicated NICs from Controller nodes for offload
- VF-LAG (SRIOV bonding) is implemented on compute nodes
- Has been tested for 6 port and 8 port deployments


ahmadfsbd:
This PR is good to go. Deployments have been tested through code.

PR #605, opened 2020-09-18
CES-11611 Configure ironic to use idrac HW type supported interfaces

himanshusingh2:


rpioso:
I resolved the merge conflicts.

rpioso (approved):
Looks good to me! Thank you, @himanshusingh2!

src/pilot/ironic.patch
```diff
@@ -1,6 +1,99 @@
...
-+++ ironic.conf.new	2020-09-08 15:21:43.975308033 -0400
-@@ -1291,7 +1291,7 @@
+--- ironic.conf	2020-09-17 02:13:45.000000000 +0000
++++ ironic.conf.new	2020-09-18 16:37:38.844775481 +0000
+@@ -37,7 +37,7 @@
+ # enumerating the "ironic.hardware.types" entrypoint. (list
+ # value)
+ #enabled_hardware_types = ipmi
+-enabled_hardware_types=idrac,ilo,ipmi,redfish
++enabled_hardware_types=idrac
```
cdearborn, line 8: Is there a reason for shutting off the ilo, ipmi, and redfish hardware types?  In the enabled_*_interfaces below, ilo and redfish are enabled, which seems odd when the hardware type is disabled.

src/pilot/ironic.patch
```diff
@@ -1,6 +1,99 @@
...
+ #default_vendor_interface = <None>
++default_vendor_interface=idrac-wsman
+ 
+ # DEPRECATED: Used if there is a formatting error when
+ # generating an exception message (a programming error). If
+@@ -701,6 +705,7 @@
+ 
+ # Seconds to wait for a response from a call. (integer value)
+ #rpc_response_timeout = 60
++rpc_response_timeout = 600
```
cdearborn, line 92: Is the change in the line above intended?

cdearborn (commented):
had a couple questions...

src/pilot/ironic.patch
rpioso: Yes, it addresses a [known issue](https://docs.openstack.org/ironic/latest/admin/drivers/idrac.html#vendor-passthru-timeout). @himanshusingh2 observed `assigne_role.py` encounter it during BIOS settings configuration.

src/pilot/ironic.patch
rpioso: Not enabling the other drivers enables us to assign hardware interface implementation values to the `default_xxx_interface` settings which are not supported by those drivers. The Conductor ensures all `default_xxx_interface`s are supported by all enabled drivers. It fails to start if it finds they are not.

We could remove the `default_xxx_interface` settings. However, that may make it more challenging to switch from WS-Man to Redfish interface implementations.

src/pilot/ironic.patch
cdearborn: Thanks!

src/pilot/ironic.patch
cdearborn: Ah, that makes sense. Thanks!

cdearborn (approved):
Looks good!

PR #606, opened 2020-09-22
Deploy edge sites seperate stacks

dpaterson:
5 port edge site add/delete functional, including routes


src/deploy/osp_deployer/director.py
```diff
@@ -28,17 +27,15 @@
...
-from auto_common.yaml_utils import OrderedDumper
 from auto_common.yaml_utils import OrderedLoader
 
-logger = logging.getLogger("osp_deployer")
+# logger = logging.getLogger("osp_deployer")
 # TODO after testing delete two logging config lines below
-# logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.DEBUG)
-# logger = logging.getLogger()
+logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.DEBUG)
+logger = logging.getLogger()
```
gaelrehault, line 38: what comment says

src/deploy/osp_deployer/settings/sample_hci.properties
```diff
@@ -24,13 +21,14 @@
...
-
         "management_ip":"192.168.110.13",
         "public_api_ip":"192.168.190.13",
-
-        "private_api_ip":"192.168.140.13"
+        "private_api_ip":"192.168.140.13",
+        "public_api_if": "enp1s0",
+        "provisioning_if": "enp2s0",
+        "management_if": "enp3s0",
+        "private_api_if": "enp4s0"
```
gaelrehault, line 31: I assume those are used for setting up routing etc, could we not do this against the bonds rather than interfaces ?

src/deploy/osp_deployer/director.py
dpaterson: Done

src/deploy/osp_deployer/settings/sample_hci.properties
dpaterson: Unfortunately you have to restart the physical interfaces for routes to take on Director VM, tried avoiding this but restarting bond, bridge and routes has no effects on routes being functional, they will show up in "ip route" but are non-functional until you ifdown/ifup the interface (enp2s0 in the case of provisioning network for example)

gaelrehault: Could we grab the interfaces name from the .properties to avoid additional/duplicate settings in ini ?

src/deploy/auto_common/ssh.py
```diff
@@ -83,8 +83,8 @@ def execute_command_tty(address, usr, pwd, command, get_pty=True):
         """
 
         try:
-            logger.debug("ssh " + usr + "@" + address + ", running : " +
-                         command)
+            logger.info("ssh " + usr + "@" + address + ", running : " +
```
gaelrehault, line 86: we should keep those at debug level

src/deploy/jinja2_templates/dell-environment-edge.j2.yaml
```diff
@@ -11,11 +11,22 @@
...
 # limitations under the License.
-
 resource_registry:
-  OS::TripleO::NodeUserData: ./first-boot.yaml
+  OS::TripleO::NodeUserData: {{node_user_data_path}}
 
 parameter_defaults:
+  NeutronPublicInterface: bond1
+  NeutronNetworkType: vlan
+  NeutronNetworkVLANRanges: physint:201:250,physext
```
gaelrehault, line 20: Should that range be hardcoded here ? Should probably be set from ini's tenant_vlan_range

src/deploy/jinja2_templates/nic-configs/5_port/nic_environment_edge.j2.yaml
```diff
@@ -18,6 +18,10 @@ resource_registry:
 {%- endfor %}
 
 parameter_defaults:
+  ControllerBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+  ComputeBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+  StorageBondInterfaceOptions: mode=802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
+  BondInterfaceSriovOptions: 802.3ad miimon=100 xmit_hash_policy=layer3+4 lacp_rate=1
```
gaelrehault, line 24: Those should be read from ini bond settings ?

src/deploy/osp_deployer/checkpoints.py
```diff
@@ -269,59 +273,123 @@ def verify_nodes_registered_in_ironic(self):
...
                           + len(setts.ceph_nodes)
                           + len(setts.computehci_nodes))
-
-        for node_type, nodes in setts.node_types_map.items():
-            logger.debug("Number of %s nodes: %s", node_type, str(len(nodes)))
-            expected_nodes += len(nodes)
-        if len(ls_nodes) != expected_nodes:
+        # hack, checking we have at least the correct number of core nodes
+        # TODO, we should get more specific for core node validation
+        if len(ls_nodes) < expected_nodes:
```
gaelrehault, line 278: Why can t we compare the exact number of expected core nodes ?

src/deploy/osp_deployer/checkpoints.py
```diff
@@ -269,59 +273,123 @@ def verify_nodes_registered_in_ironic(self):
...
         if "No such file or directory" in re[0]:
-            raise AssertionError(
-                "Director & Undercloud did not install properly, "
-                "check /pilot/install-director.log for details")
+            _error = AssertionError("Director & Undercloud did not install "
+                                    "properly, no ~/stackrc found, check "
+                                    "/pilot/install-director.log "
+                                    "for details")
+
+            return True, _error
```
gaelrehault, line 378: Why not raise the error here ?

src/deploy/osp_deployer/deployer.py
```diff
@@ -82,16 +97,124 @@ def get_settings():
...
+    return _is_deploy_overcloud
+
+
+def get_is_deploy_undercloud_and_overcloud():
+    _is_deploy_undercloud = False
+    _input = input("Deploy the undercloud and overcloud prior to "
+                   "deploying edge site(s), (y/N)? ")
+    if _input.lower().strip() == "y":
+        _is_deploy_undercloud = True
+    return _is_deploy_undercloud
```
gaelrehault, line 169: Not a fan of asking for user input .. could we not just add a --edge(s)_only param to be consistant with existing options ?

src/deploy/osp_deployer/deployer.py
```diff
@@ -82,16 +97,124 @@ def get_settings():
...
+    for node_type in settings.node_types:
+        logger.info("  {}".format(node_type))
+    for node_type, node_type_data in settings.node_type_data_map.items():
+        logger.info("\nEdge site {} "
+                    "metadata:"
+                    "\n{}\n".format(node_type, pformat(node_type_data)))
+        nodes = settings.node_types_map[node_type]
+        logger.info("\nEdge site: {} nodes:".format(node_type))
+        for node in nodes:
+            logger.info("\n{}\n".format(pformat(node.__dict__)))
```
gaelrehault, line 216: We should clarify this is based on deployment settings, not what might currently be deployed

src/deploy/osp_deployer/deployer.py
```diff
@@ -114,126 +260,105 @@ def deploy():
...
+        if (is_deploy_edge_site and _is_undercloud_failed):
+            _is_deploy_both = get_is_deploy_undercloud_and_overcloud()
+            if _is_deploy_both:
+                deploy_undercloud(settings, sah_node, tester, director_vm)
+                # recheck undercloud
+                _is_uc_failed, _err_uc = tester.verify_undercloud_installed()
+                if not _is_uc_failed:
+                    deploy_overcloud(director_vm)
+                else:
+                    raise _err_uc
```
gaelrehault, line 314: I think that s overkill, if undercloud failed,  likelihood is it ll faill for the same reason again

src/deploy/osp_deployer/deployer.py
```diff
@@ -114,126 +260,105 @@ def deploy():
...
+                _is_oc_failed, _err_oc = tester.verify_overcloud_deployed()
+                if not _is_oc_failed:
+                    deploy_edge(director_vm, node_type)
+                else:
+                    raise _err_oc
+            else:
+                logger.warn("Attempted to deploy edge site(s) but the "
+                            "undercloud and overcloud are not deployed "
+                            "and the user declined to deploy "
+                            "the them first, follwed by edge site(s). "
```
gaelrehault, line 325: the them > them

src/deploy/osp_deployer/settings/sample_hci.properties
gaelrehault: Sorry i was thinking this was for the SAH interfaces. If tis is director vm, let s not make them a settings, it s a pre set list of interfaces that ll always be enp1s0...enp4s0, should not be "configurable"

src/pilot/deploy-overcloud.py
```diff
@@ -623,6 +624,7 @@ def main():
               "environments/ceph-ansible/ceph-rgw.yaml" \
               " {}" \
               " --libvirt-type kvm" \
+              " --no-cleanup" \
```
gaelrehault, line 627: no issue with having this, but any reason ?

src/deploy/jinja2_templates/dell-environment-edge.j2.yaml
dpaterson: The following are hard-coded in pilot/templates/dell-environment.yaml and never modified by automation
  NeutronPublicInterface: bond1
  NeutronNetworkType: vlan
  NeutronBridgeMappings: physint:br-tenant,physext:br-ex

The attribute below is updated before the overcloud deployment in pilot/dell_nfv.py,  I missed that we use that module for all deployments, not just NFV type deployments.
NeutronNetworkVLANRanges: physint:201:250,physext

I'll populate those variables with Jinja2, we may want to retrofit dell_nfv.py to also set NeutronNetworkType and NeutronPublicInterface.

src/deploy/osp_deployer/checkpoints.py
dpaterson: It should be done, I was pushing it off as of now we only do openstack baremetal node list and count how many there are.  The validation will need to be updated to examine each baremetal node's properties, specifically:
 node['capabilities']['service_tag'] and validate the overcloud nodes defined in .properties have exact service tag matches in the ironic database.
I'll add that functionality and update the PR

src/deploy/jinja2_templates/dell-environment-edge.j2.yaml
dpaterson: +1, will do

src/deploy/osp_deployer/deployer.py
dpaterson: Right now when all you pass in is -s the automation does a full deployment, how should it behave if only -s and --edge_site args are provided,  just assume we are doing a full deployment + an edge site?  Seems dangerous,  the alternative is we make --edge_site mutually exclusive where it fails fast if there is no undercloud and overcloud to add an edge site to.  The detractor being you have to run two commands instead of one, #1 to deploy director/overcloud, #2 to deploy the edge site (--edge_site arg).  I'm all for implementing it this way but wasn't sure how it would go over so I added the prompt.  Please advise.

src/deploy/osp_deployer/deployer.py
dpaterson: +1, good idea, this brings up a good point, what if user runs deployer.py ... --edge-site edge-compute-denver twice?  Do we simply delete the stack and redeploy the edge site?
Seems reasonable, please confirm this should be the intended behavior.

I will also add deployment state to --list_edge_sites output.

src/deploy/osp_deployer/deployer.py
dpaterson: This is checking that the undercloud deployment did not fail and then deploy the overcloud if it doesn't fail.  I believe we are taking out the prompt all together (see comment above) so this will be a mute point.

src/pilot/deploy-overcloud.py
dpaterson: Ah, I believe this was just for debugging, but that said, does it hurt to leave that behind for post-mortem, could be useful for debugging failed deployment?  Please advise.

src/deploy/auto_common/ssh.py
dpaterson: +1, Done

src/deploy/osp_deployer/settings/sample_hci.properties
dpaterson: Will move to constants and out of ini.

src/deploy/osp_deployer/deployer.py
gaelrehault: _hat if user runs deployer.py ... --edge-site edge-compute-denver twice? Do we simply delete the stack and redeploy the edge site?_
I would say yes, no different that doing an overcloud_only if overcloud already deployed, but would like to hear from @heydog15 and rest of the team also

src/pilot/deploy-overcloud.py
gaelrehault: happy to leave that in if usefull for debugging

src/deploy/jinja2_templates/dell-environment-edge.j2.yaml
dpaterson: Updated to get attribute from tenant_vlan_range, will be in next patch set.

src/deploy/jinja2_templates/nic-configs/5_port/nic_environment_edge.j2.yaml
dpaterson: Done, using config.get_curated_nics_settings() to fetch bonding interface options,  change in next patch set.

src/deploy/osp_deployer/checkpoints.py
dpaterson: Done, will be reflected in next patch set.

src/deploy/osp_deployer/settings/sample_hci.properties
dpaterson: Done

src/deploy/osp_deployer/deployer.py
dpaterson: Not relevant anymore as logic has been refactored.

src/pilot/deploy-overcloud.py
dpaterson: Let's keep it then

src/deploy/osp_deployer/deployer.py
dpaterson: Done, change will be in next patch set

PR #607, opened 2020-09-25
CES-11261 updated assign_role ...

gaelrehault:
- to handle missing attribute prior to raid/bios config ( causing errors during cleaning )
- removed unused code.

### DIFF
--- a/src/pilot/assign_role.py
+++ b/src/pilot/assign_role.py
@@ -776,6 +776,8 @@ def configure_raid(ironic_client, node_uuid, super_role, os_volume_size_gb,
     ironic_client.node.wait_for_provision_state(node_uuid, 'manageable')
     LOG.info("Completed deletion of the existing RAID configuration")
 
+    # Work around the bugs in the ironic DRAC driver's RAID clean steps.
+
     target_raid_config = define_target_raid_config(
         super_role, drac_client)
 
@@ -788,6 +790,7 @@ def configure_raid(ironic_client, node_uuid, super_role, os_volume_size_gb,
 
     # Set the target RAID configuration on the ironic node.
     ironic_client.node.set_target_raid_config(node_uuid, target_raid_config)
+
 
     LOG.info("Applying the new RAID configuration")
     clean_steps = [{'interface': 'raid', 'step': 'create_configuration'}]
@@ -843,19 +846,16 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
     if role_index.index:
         role = "node:{}-{}".format(flavor, role_index.index)
 
-    if 'capabilities' in node.properties:
-        value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
-        LOG.info(str(node.properties))
-        LOG.info(str(value))
-    else:
-        value = "{},boot_option:local".format(role)
-        LOG.info("..!")
+    value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
+    LOG.info(str(node.properties))
+    LOG.info(str(value))
 
     patch = [{'op': 'add',
               'value': value,
               'path': '/properties/capabilities'}]
     ironic_client.node.update(node_uuid, patch)
     LOG.info(str(patch))
+
 
 
 def generate_osd_config(ip_mac_service_tag, drac_client):
@@ -964,7 +964,6 @@ def generate_osd_config(ip_mac_service_tag, drac_client):
         fcntl.flock(stream, fcntl.LOCK_UN)
         stream.close()
 
-
 def get_drives(drac_client):
 
     spinners = []
@@ -1033,7 +1032,6 @@ def get_drives(drac_client):
 
     return spinners, ssds, nvme_drives
 
-
 def generate_osd_config_without_journals(controllers, drives):
 
     osd_config = {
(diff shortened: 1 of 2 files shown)


heydog15:
This code has been deployed and tested successfully on a few racks.

src/pilot/assign_role.py
```diff
@@ -778,32 +778,6 @@ def configure_raid(ironic_client, node_uuid, super_role, os_volume_size_gb,
 
     # Work around the bugs in the ironic DRAC driver's RAID clean steps.
```
rpioso, line 780: Let us delete these two (2) lines, too.

src/pilot/assign_role.py
```diff
@@ -817,35 +791,6 @@ def configure_raid(ironic_client, node_uuid, super_role, os_volume_size_gb,
...
-        node_uuid, ironic_client, drac_client, 'RAID',
-        controllers_to_physical_disk_ids)
-
-    if succeeded:
-        LOG.info("Completed converting physical disks configured to back RAID "
-                 "logical disks to RAID mode")
-    else:
-        LOG.critical("Attempt to convert physical disks configured to back "
-                     "RAID logical disks to RAID mode failed")
-        return False
```
rpioso, line 794: Please delete this blank line, too. My Python-aware Eclipse PyDev editor complains loudly about extra lines.

rpioso (commented):
I have one (1) major comment and a couple of minor ones.

Both @cdearborn and I believe the code in the `assign_role()` function which ensures `boot_mode:uefi` is in the `capabilities` property on the ironic bare metal node object should be unconditionally executed. Presently, it is conditioned on the presence of `capabilities` in `node.properties`. Any reason why that is?

Our gut is that it happens to work, because by the time it runs, the node has already been in-band introspected and has `capabilities`. Fortunate :-)

The block of code I am referring to is in `assign_role.py`, lines 849-855. For your convenience, I have copied it below.

    if 'capabilities' in node.properties:
        value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
        LOG.info(str(node.properties))
        LOG.info(str(value))
    else:
        value = "{},boot_option:local".format(role)
        LOG.info("..!")

If necessary, this code could be merged as-is and a follow-up PR submitted to address my comments. I would prefer they be addressed now. Your call, though.

gaelrehault:
Addressed the minor comments. 

In regard to "_Presently, it is conditioned on the presence of capabilities in node.properties. Any reason why that is?_", I can only point you to the following PR, reviewed and merge by yourself nearly 2 years ago , that introduced the condition in the first place : 
https://github.com/dsp-jetpack/JetPack/pull/183 

If you feel this should now be questioned & removed, i m happy to take it out, but this isn't something that was recently introduced as you can see from the above; 
Given the time constraint we re under, I would rather address it at a later stage, but leaving that decision up to @heydog15 .

gaelrehault:
Richard, addressed the minor issues. 
In relation to the capabilities, would you mind clarifying a little more ? what i m understanding you re saying, is that capabilities will only be available if introspection ran in-band, but not out of band ? If that s the case we should add the property in both cases ?

heydog15:
We need both in-band and out of band either way.  Until we can do out of band NFV stuff, we need both.

rpioso:
> Richard, addressed the minor issues.
> In relation to the capabilities, would you mind clarifying a little more ? what i m understanding you re saying, is that capabilities will only be available if introspection ran in-band, but not out of band ? If that s the case we should add the property in both cases ?

Let me try again. The `if` statement is not needed. Just do it! (tm) without the `if` statement and `else` clause.

    value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
    LOG.info(str(node.properties))
    LOG.info(str(value))

Further clarifying, `capabilities` is available if in-band introspection (IB) **or** out-of-band (OOB) inspection is performed before `assign_role.py` is executed, which I believe it is. Agree? If so, that is why the code has been working as-is. If the order were reversed, it would not work. So, the current code is not robust, confusing, and difficult to maintain.

I understand IB is presently used by JetStrem, because of the need for NUMA topology information, which OOB does not presently offer. Should that information be added to OOB and JetStream changed to use OOB, my suggestion would continue to work.

gaelrehault:
Removed condition , and added missing commit required

src/pilot/assign_role.py
```diff
@@ -901,26 +846,16 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index, os_volume_size_gb,
...
-    if 'capabilities' in node.properties:
-        value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
-        LOG.info(str(node.properties))
-        LOG.info(str(value))
-    else:
-        value = "{},boot_option:local".format(role)
-        LOG.info("..!")
+    value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
+    LOG.info(str(node.properties))
+    LOG.info(str(value))
```
rpioso, line 851: As a follow-up to this pull request, let's make this work when out-of-band (OOB) inspection or in-band (IB) introspection has not been performed before `assign_role.py` is run. That workflow is used by the Ironic Team when integration testing their changes. In that case, the `capabilities` property may not be present and line 849 above would cause an exception to be raised.

The new implementation could be something like the following:

    if 'capabilities' in node.properties:
        value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
    else:
         value = "{},boot_mode:uefi,boot_option:local".format(role)

    LOG.info(str(node.properties))
    LOG.info(str(value))

rpioso (commented):
LGTM! Thank you, @gaelrehault and @himanshusingh2!

I suggest we create a follow-up pull request to address a shortcoming which the Ironic Team would likely encounter when using `assign_role.py` as an integration test tool. It does not affect JetStream nor automated deployment.

PR #610, opened 2020-10-01
Fixing CES-11633 Sanity does not support for Unity in JS-16.1 release

rajinir:
The support for sanity test for manila shares was not migrated
from JS13.3


rajinir:
This can be merged. Tested that it deployed successfully and created shares and passed sanity

rajinir:
Rebased.

PR #615, opened 2020-10-07
updated assign role - JS-16.1 

gaelrehault:


rpioso (changes requested):
I would prefer this change be added to either `experimental-upstream-ironic-JS-16.1-00` or `JS-16.1` and then merged.

Otherwise, we are setting ourselves up for merge migraines.

heydog15:
Looks good.

PR #617, opened 2020-10-07
fixing issue with subscription manager not available post director no…

gaelrehault:
…de install  (JS-13.3 locked install )


heydog15:
Looks good.

PR #622, opened 2020-10-12
Fix assign_role.py modification of node capabilities

himanshusingh2:
* follow-up pull request to CES-11261 Update assign_role.py.
* boot_mode:uefi is in the capabilities property on the ironic bare metal node object will be unconditionally executed.


gaelrehault (approved):

src/pilot/assign_role.py
```diff
@@ -846,7 +846,11 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
     if role_index.index:
         role = "node:{}-{}".format(flavor, role_index.index)
 
-    value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
+    if 'capabilities' in node.properties:
+        value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
```
rpioso, line 850: What happens if `boot_mode` and/or `boot_option` are already in `capabilities`? It seems to me they should not be added.

This could occur when OOB inspection or IB introspection is performed before `assign_role.py`. Our `idrac` driver's OOB inspection, both WS-Man and Redfish, add `boot_mode` to `capabilities` if it is not already present and IB introspection can be configured to do the same, although its default setting is not to do that.

rpioso (changes requested):
Thank you, @himanshusingh2! I added a couple of requests in-line.

src/pilot/assign_role.py
```diff
@@ -846,7 +846,11 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
     if role_index.index:
         role = "node:{}-{}".format(flavor, role_index.index)
 
-    value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
+    if 'capabilities' in node.properties:
```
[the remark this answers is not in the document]
himanshusingh2: `capabilities` is available if in-band introspection (IB) or out-of-band (OOB) inspection is performed before `assign_role.py` is executed, This `if` `else` condition will make this work in a controlled manner when in-band (IB) introspection or out-of-band (OOB) inspection has not been performed before `assign_role.py` is run and add `boot_mode` and `boot_option` to the node properties.

src/pilot/assign_role.py
himanshusingh2: I believe this will overwrite it as `boot_mode:uefi` and `boot_option:local`

src/pilot/assign_role.py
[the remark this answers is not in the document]
ajya: clarified requirements:

If there are no capabilities then add capabilities with role, boot_mode:uefi and boot_option: local (as in line 857 already)
If there are capabilities, then
- add role
- for boot_mode and boot_option update or add it so that there are boot_mode:uefi, boot_option:local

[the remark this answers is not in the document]
rpioso: Please consider this approach: https://github.com/openstack/ironic/blob/349bf2d5bcb38b409fe01aa78cc281f016199efc/ironic/common/utils.py#L402.

src/pilot/assign_role.py
```diff
@@ -846,7 +846,30 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
...
+        current_capabilities = node.properties.get('capabilities')
+        try:
+            cap_dict = dict(x.split(':', 1)
+                            for x in current_capabilities.split(','))
+        except ValueError:
+        # ValueError can occur if capabilities values not properly
+        # defined like capabilities='boot_mode:uefi,boot_option'.
+            raise ValueError(
+                _("Invalid capabilities string '%s'.") % current_capabilities)
```
ajya, line 863: no need for underscore here

src/pilot/assign_role.py
```diff
@@ -846,7 +846,30 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
...
+    # inspection or in-band introspection has been performed.
+    if 'capabilities' in node.properties:
+        current_capabilities = node.properties.get('capabilities')
+        try:
+            cap_dict = dict(x.split(':', 1)
+                            for x in current_capabilities.split(','))
+        except ValueError:
+        # ValueError can occur if capabilities values not properly
+        # defined like capabilities='boot_mode:uefi,boot_option'.
+            raise ValueError(
```
ajya, line 862: @gaelrehault is it necessary to raise error here or rather LOG error and return? If raising error, is it OK to raise `ValueError` or should it be different error?

src/pilot/assign_role.py
```diff
@@ -846,7 +846,30 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
     if role_index.index:
         role = "node:{}-{}".format(flavor, role_index.index)
 
-    value = "{},{},boot_mode:uefi,boot_option:local".format(role, node.properties['capabilities'])
+    # capabilities will be available if in-band introspection or
+    # out-of-band inspection is performed before
+    # assign_role.py is executed.
+    # if else statement needed to run this script before out-of-band
```
ajya, line 852: Let's replace "if else statement" with "This conditional handling is" for better readability

ajya (changes requested):
Thanks for updates, some comments inline.

src/pilot/assign_role.py
```diff
@@ -846,7 +846,30 @@ def assign_role(ip_mac_service_tag, node_uuid, role_index,
...
+    if 'capabilities' in node.properties:
+        current_capabilities = node.properties.get('capabilities')
+        try:
+            cap_dict = dict(x.split(':', 1)
+                            for x in current_capabilities.split(','))
+        except ValueError:
+        # ValueError can occur if capabilities values not properly
+        # defined like capabilities='boot_mode:uefi,boot_option'.
+            raise ValueError(
+                    ("Invalid capabilities string '%s'.") % current_capabilities)
```
ajya, line 863: 1. fix indentation to be 4 spaces from previous line
2. remove () around `("Invalid capabilities string '%s'.")` - not necessary after removing `_`

src/pilot/assign_role.py
rpioso: @gaelrehault, this is a friendly reminder that we would appreciate your input on this.

ajya (approved):
thanks for updates

src/pilot/assign_role.py
gaelrehault: If we have invalid capabilities i think it s reasonable to raise an error yes.
i have no opinion on the exception type used.

rpioso (approved):
Looks great to me! Thank you, @himanshusingh2!

PR #627, opened 2020-10-14
CES-11670 

gaelrehault:
.fix issue assigning root device


heydog15:
Looks good.  Make sure it gets re-tested on R7 and R60

PR #630, opened 2020-10-16
Fix CES-11672 - bin_physical_disks_by_size_gb() returns floating point

mahendraKamble:
bin_physical_disks_by_size_gb method returns floating type value. 
It should return an int value.


gaelrehault (approved):

sonali85:
Hi All,
I have tested this fix against 14G and 13G . Let me know if we required any changes .

rpioso (approved):
Looks great to me! Thank you, @sonali85 and @mahendraKamble!

PR #633, opened 2020-10-21
Add 8 port DPDK/SRIOV edge support

dpaterson:


src/deploy/osp_deployer/deployer.py
```diff
@@ -241,7 +241,9 @@ def deploy():
         settings, args = get_settings()
         director_vm = Director()
         sah_node = Sah()
-
+        # director_vm.render_and_upload_compute_edge("edge-spain")
+        # director_vm.render_and_upload_sriov_edge("edge-spain")
+        # os._exit(0)
```
gaelrehault, line 246: should remove the above

dpaterson: Done

PR #647, opened 2020-11-12
Support multiple RAID controllers for compute role

mpardhi23:
Hey @rpioso,
This change is to add support of multiple controller for night's watch server. 
I have implemented code for support of multiple controller in compute node which is currently tested against the available controller combination i.e. BOSS + PERC H740 . Able to create RAID1 and RAID10 correspondingly when BOSS + PERC controller found.
Testing for other scenarios is pending including deployment run. Currently I am testing this new changes and also run regression testing as well.
Note: I will enhance the code for new implementation as well, once I complete all test functionalities.


mpardhi23:
Hi Guys,
I have tested multiple controller code by running OSP deployment by configuring night's watch server which is having BOSS + PERC combination as compute node in properties file . Assign_role executed successfully on all nodes and able to create RAID1 and RAID10 correspondingly on night's watch server.  Deployment is successful till overcloud installation.

rpioso:
@dpaterson, do you still plan to review this pull request? If not, perhaps @gaelrehault or @heydog15 could.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,120 @@ def get_raid_controller_ids(drac_client):
     return raid_controller_ids
 
 
-def define_controller_logical_disks(drac_client, raid_controller_ids):
-    if len(raid_controller_ids) > 1:
-        LOG.critical(
-            "Found more than one RAID controller:\n  {}".format(
-                "\n  ".join(raid_controller_ids)))
-        return None
+def define_controller_logical_disks(drac_client, raid_controllers):
```
dpaterson, line 299: Is it cleaner to only pass in drac_client and fetch raid_controllers in this method as we do with all_physical_disks...?

src/pilot/assign_role.py
```diff
@@ -296,48 +296,120 @@ def get_raid_controller_ids(drac_client):
...
+            if disk.controller in raid_cntlr_physical_disks:
+                raid_cntlr_physical_disks[disk.controller].append(disk)
+            else:
+                raid_cntlr_physical_disks[disk.controller] = [disk]
 
-    # None indicates an error occurred.
-    if raid_10_logical_disk is None:
+    boss_controller = [cntrl for cntrl in raid_controllers
+                       if drac_client.is_boss_controller(cntrl)]
+
```
dpaterson, line 313: Perhaps add new function _fetch_boss_controller is cleaner than this construct, also doesn't this return a array so for it to be correct it would be:
[cntrl for cntrl in raid_controller if drac_client.is_boss_controller(cntrl)][0] ?

src/pilot/assign_role.py
```diff
@@ -296,48 +296,120 @@ def get_raid_controller_ids(drac_client):
...
+    if boss_controller:
+        if check_cntlr_physical_disks_len(
+           raid_cntlr_physical_disks[boss_controller[0]]):
```
dpaterson, line 316: Not a fan of boss_controller being a list, there can only be one boss card per system so we should just have single method _fetch_boss_card or something like that, that returns None or the boss card, not a list.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,120 @@ def get_raid_controller_ids(drac_client):
...
+                raid_cntlr_physical_disks[disk.controller] = [disk]
 
-    raid_10_logical_disk = define_single_raid_10_logical_disk(
-        drac_client, raid_controller_ids[0])
+    boss_controller = [cntrl for cntrl in raid_controllers
+                       if drac_client.is_boss_controller(cntrl)]
+
+    if boss_controller:
+        if check_cntlr_physical_disks_len(
+           raid_cntlr_physical_disks[boss_controller[0]]):
```
dpaterson, line 375: Same comment as above, only one boss card per box is possible.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,120 @@ def get_raid_controller_ids(drac_client):
...
 
 
-def define_compute_logical_disks(drac_client, raid_controller_ids):
-    if len(raid_controller_ids) > 1:
-        LOG.critical(
-            "Found more than one RAID controller:\n  {}".format(
-                "\n  ".join(raid_controller_ids)))
-        return None
+def define_compute_logical_disks(drac_client, raid_controllers):
+    all_physical_disks = drac_client.list_physical_disks()
```
dpaterson, line 359: Same comment as above, use drac_client to fetch controllers just like we fetch all_physical_disks

dpaterson (changes requested):
Couple of small changes would make this cleaner but I don't see anything wrong with logic at first sight,  has this been tested against real hardware?

src/pilot/assign_role.py
mpardhi23: If we fetch raid_controllers in this method then I need to change the existing code , where this function get invoked in  define_target_raid_config function. In this function we already fetched raid_controllers.
line no. 263 : logical_disks = define_controller_logical_disks(drac_client, raid_controller_ids) 
I think we can keep it as is.

mpardhi23:
> Couple of small changes would make this cleaner but I don't see anything wrong with logic at first sight, has this been tested against real hardware?

Yes I have tested these changes against night's watch server XE2420.

src/pilot/assign_role.py
mpardhi23: Agree. I will add new function which fetch the boss controller and return None or boss card.

src/pilot/assign_role.py
mpardhi23: Okay , will do this changes and push the latest changes.

src/pilot/assign_role.py
mpardhi23: sure

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: I think we can keep it as is.

dpaterson (approved):
Thanks for implementing suggested changes

mpardhi23:
> Thanks for implementing suggested changes

Thanks David!

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
-            "Found more than one RAID controller:\n  {}".format(
-                "\n  ".join(raid_controller_ids)))
-        return None
+def define_controller_logical_disks(drac_client, raid_controllers):
+    all_physical_disks = drac_client.list_physical_disks()
 
-    raid_10_logical_disk = define_single_raid_10_logical_disk(
-        drac_client, raid_controller_ids[0])
+    # Get the drives controlled by the RAID controller
+    raid_cntlr_physical_disks = {}
```
ajya, line 303: this can be shortened by using defaultdict(list)

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
+        if check_cntlr_physical_disks_len(
+           raid_cntlr_physical_disks[boss_controller]):
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[boss_controller],
+                drac_client, boss_controller)
+        else:
+            LOG.critical("The BOSS card has only 1 SSD. "
+                         "2 SSDs are needed to configure a RAID 1")
+            return None
+    else:
```
ajya, line 322: no need for else here - RAID 10 should be executed in both cases - when there is boss controller and when there is not. The remaining code below that tries to create RAID10 again can be removed.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
+def define_controller_logical_disks(drac_client, raid_controllers):
+    all_physical_disks = drac_client.list_physical_disks()
 
-    raid_10_logical_disk = define_single_raid_10_logical_disk(
-        drac_client, raid_controller_ids[0])
+    # Get the drives controlled by the RAID controller
+    raid_cntlr_physical_disks = {}
+    for disk in all_physical_disks:
+        if disk.controller in raid_controllers:
+            if disk.controller in raid_cntlr_physical_disks:
```
ajya, line 306: this can be shortened by using defaultdict(list) for raid_cntlr_physical_disks

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
+           raid_cntlr_physical_disks[boss_controller]):
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[boss_controller],
+                drac_client, boss_controller)
+        else:
+            LOG.critical("The BOSS card has only 1 SSD. "
+                         "2 SSDs are needed to configure a RAID 1")
+            return None
+    else:
+        os_logical_disk = define_single_raid_10_logical_disk(
```
ajya, line 323: update define_single_raid_10_logical_disk to have optional is_root_volume param with default True (as it is used now).

Then based on scenario: 
1) if there is boss controller, pass is_root_volume=False
2) is there is no boss controller, pass is_root_volume=True

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
     return logical_disks
 
 
-def define_compute_logical_disks(drac_client, raid_controller_ids):
-    if len(raid_controller_ids) > 1:
-        LOG.critical(
-            "Found more than one RAID controller:\n  {}".format(
-                "\n  ".join(raid_controller_ids)))
-        return None
+def define_compute_logical_disks(drac_client, raid_controllers):
```
ajya, line 356: this method looks very similar to define_controller_logical_disks and too much of code is duplicating, need to eliminate duplication. Two of the options:

1) merge the method into one, named e.g. define_controller_or_compute_logical_disks and update where they both are called.
2) introduce private helper method, e.g., _define_controller_or_compute_logical_disks and call it from these 2 methods. This way don't have to change all the places where these 2 are called now.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
+            if disk.controller in raid_cntlr_physical_disks:
+                raid_cntlr_physical_disks[disk.controller].append(disk)
+            else:
+                raid_cntlr_physical_disks[disk.controller] = [disk]
+
+    boss_controller = _fetch_boss_controller(drac_client, raid_controllers)
+    if boss_controller:
+        if check_cntlr_physical_disks_len(
+           raid_cntlr_physical_disks[boss_controller]):
+            os_logical_disk = define_storage_operating_system_logical_disk(
```
rpioso, line 315: Having this function, which is used for the control and compute roles, call another function which is named for the storage role can be confusing.

Let us make the following function name changes:

1. `define_storage_operating_system_logical_disk` to `define_operating_system_logical_disk`
2. `find_physical_disks_for_storage_os` to `find_physical_disks_for_os`

src/pilot/assign_role.py
rpioso: @mpardhi23, you may find https://stackoverflow.com/questions/26367812/appending-to-list-in-python-dictionary helpful.

src/pilot/assign_role.py
rpioso: FWIW, I prefer option 2. That option makes it less likely that a programmer will introduce an error by not updating all of the callers. Since this script does not have any unit tests, such errors can very easily go undetected.

rpioso (changes requested):
Thank you, @mpardhi23! I have left several in-line comments which offer suggestions for improvement.

src/pilot/assign_role.py
mpardhi23: yes removed else and remaining below code from here.

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Added _define_controller_or_compute_logical_disks method which defines the compute and controller logical disk.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
     return raid_controller_ids
 
 
-def define_controller_logical_disks(drac_client, raid_controller_ids):
-    if len(raid_controller_ids) > 1:
-        LOG.critical(
-            "Found more than one RAID controller:\n  {}".format(
-                "\n  ".join(raid_controller_ids)))
-        return None
+def define_controller_logical_disks(drac_client, raid_controllers):
```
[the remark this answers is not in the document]
mpardhi23: Thanks Richard for review comments, replaced raid_controllers to raid_controller_ids. as its listing only controller ids.

src/pilot/assign_role.py
mpardhi23: Done

mpardhi23:
Hi Richard and Aija,
I have addressed all the review comments you mentioned. And also tested those changes with the below controllers available on server.
1. H740P 
2. BOSS + H740P
Need to test it against another controllers which are available on another servers (different stamps)

src/pilot/assign_role.py
```diff
@@ -297,52 +297,52 @@ def get_raid_controller_ids(drac_client):
...
+        if disk.controller in raid_controller_ids:
+            raid_cntlr_physical_disks[disk.controller].append(disk)
 
-def define_compute_logical_disks(drac_client, raid_controller_ids):
-    if len(raid_controller_ids) > 1:
-        LOG.critical(
-            "Found more than one RAID controller:\n  {}".format(
-                "\n  ".join(raid_controller_ids)))
-        return None
+    is_root_volume=True
```
ajya, line 318: let's use `boss_controller` to track if RAID 10 will be root volume or not. This way can avoid introducing another variable. After removing `is_root_volume`, on line 335 use `not boss_controller`.

src/pilot/assign_role.py
```diff
@@ -297,52 +297,52 @@ def get_raid_controller_ids(drac_client):
...
+        drac_client, raid_controller_ids[0], is_root_volume)
 
-    # None indicates an error occurred.
-    if raid_10_logical_disk is None:
-        return None
 
-    logical_disks = list()
-
-    # Add the disk to the list only if it is not empty.
     if raid_10_logical_disk:
```
ajya, line 338: if there is no RAID 10 created, let's return `None` as it is mandatory to have it on compute and controller nodes.

src/pilot/assign_role.py
```diff
@@ -405,6 +405,13 @@ def check_cntlr_physical_disks_len(cntrl_physical_disks):
         return True
 
 
+def _fetch_boss_controller(drac_client, raid_controllers):
+    boss_controller = [cntrl for cntrl in raid_controllers
+                       if drac_client.is_boss_controller(cntrl)]
+    if boss_controller:
+        return boss_controller[0]
```
ajya, line 412: @gaelrehault and @heydog15, what should be done if there is more than 1 BOSS found in the system? Is it OK to return the first one?

src/pilot/assign_role.py
```diff
@@ -478,10 +485,10 @@ def define_storage_logical_disks(drac_client, raid_controllers):
     return logical_disks
 
 
-def define_storage_operating_system_logical_disk(physical_disks, drac_client,
+def define_operating_system_logical_disk(physical_disks, drac_client,
                                                  raid_controller_name):
     (os_logical_disk_size_gb,
-     os_physical_disk_names) = find_physical_disks_for_storage_os(
+     os_physical_disk_names) = find_physical_disks_for_os(
         physical_disks)
```
ajya, line 492: can try to wrap line 491 and 492 into one if it not as long anymore

ajya (changes requested):
thanks, looks good, some more comments added.

About testing - should test for case when there is only BOSS Controller. In stamps storage node can be used where there is BOSS controller and HBA330. As HBA330 is non-RAID controller, this system will be treated as having only BOSS controller.

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: Done

src/pilot/assign_role.py
mpardhi23: removed _is_root_volume_ and used not _boss_controller_.

src/pilot/assign_role.py
```diff
@@ -296,48 +296,116 @@ def get_raid_controller_ids(drac_client):
...
+    if boss_controller:
+        if check_cntlr_physical_disks_len(
+           raid_cntlr_physical_disks[boss_controller]):
+            os_logical_disk = define_storage_operating_system_logical_disk(
+                raid_cntlr_physical_disks[boss_controller],
+                drac_client, boss_controller)
+        else:
+            LOG.critical("The BOSS card has only 1 SSD. "
+                         "2 SSDs are needed to configure a RAID 1")
+            return None
```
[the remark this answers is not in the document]
gaelrehault: Correct, there should always be x2 disks available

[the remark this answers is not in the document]
mpardhi23: Thank you @gaelrehault  for this information!

mpardhi23:
> thanks, looks good, some more comments added.
> 
> About testing - should test for case when there is only BOSS Controller. In stamps storage node can be used where there is BOSS controller and HBA330. As HBA330 is non-RAID controller, this system will be treated as having only BOSS controller.

Tested this code for the case when there is only BOSS controller , it is failing with this changes , need to add one condition  where it check controller should not be BOSS controller before below line. 
raid_10_logical_disk = define_single_raid_10_logical_disk(drac_client, raid_controller_ids[0], not boss_controller)

src/pilot/assign_role.py
mpardhi23: @gaelrehault, confirmed that 2 BOSS cards is not valid configuration.

mpardhi23:
Hi @rpioso and @ajya,
I have updated latest changes which handles the warning part when more than one RAID controller and BOSS card found on server and selecting the first one from the list.

ajya (approved):
Thanks for updates

jproque-dell (approved):
Looks good to me. Nothing to add to what has been modified already.

mpardhi23:
> Looks good to me. Nothing to add to what has been modified already.

Thank you JP for the review and approval !

rpioso (approved):
Looks great to me! Thank you, Monica.

PR #649, opened 2020-11-16
Both sriov and dpdk

dpaterson:
This pull requests enables both SRIOV and DPDK to both be utilized simultaneously at the same edge site, aslo it addresses where you should be able to enable NUMA + Hugepages without SRIOV or DPDK (nfv_type=None)


src/deploy/osp_deployer/deployer.py
```diff
@@ -241,13 +241,15 @@ def deploy():
         settings, args = get_settings()
         director_vm = Director()
         sah_node = Sah()
-
+        # director_vm.setup_templates_edge("edge-spain")
+        # os._exit(0)
```
gaelrehault, line 245: above to remove

dpaterson: Done

PR #650, opened 2020-11-20
powerflex related updates

gaelrehault:


src/deploy/osp_deployer/deployer.py
```diff
@@ -233,6 +233,7 @@ def validate_edge_sites_in_settings(args, setts):
 
 
 def deploy_powerflex_gw(setts, sah_node, tester, powerflexgw_vm):
+    settings = setts
```
jproque-dell, line 236: Is this line needed? we're passing settings as setts, and is called as such afterward.

src/deploy/osp_deployer/deployer.py
```diff
@@ -262,6 +263,7 @@ def deploy_powerflex_gw(setts, sah_node, tester, powerflexgw_vm):
 
 
 def deploy_powerflex_mgmt(setts, sah_node, tester, powerflexmgmt_vm):
+    settings = setts
```
jproque-dell, line 266: same

jproque-dell (approved):

src/deploy/osp_deployer/deployer.py
gaelrehault: see a couple of lines further down : 
settings.powerflexgw_vm.root_password
is called a few times

jproque-dell: should be setts.powerflexgw_vm.root_password. I figure I missed to change it.

gaelrehault: done

src/deploy/osp_deployer/deployer.py
gaelrehault: done

PR #651, opened 2020-12-01
Fixed kernel-args for sriov on edge

ahmadfsbd:
Re-organized a couple of lines so that kernel args gets picked up in case of sriov.


dpaterson (approved):
LGTM, thanks

PR #654, opened 2020-12-14
Update Copyrights in JetPack to 2021 - 11700

CalvinSuma:
Updated the copyright date from x-2020 to x-2021

### DIFF
--- a/src/mgmt/director.cfg
+++ b/src/mgmt/director.cfg
@@ -1,4 +1,4 @@
-# (c) 2014-2021 Dell
+# (c) 2014-2021 Dell Inc. or its subsidiaries.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/mgmt/director.cfg
+++ b/src/mgmt/director.cfg
@@ -1,4 +1,4 @@
-# (c) 2014-2020 Dell
+# (c) 2014-2021 Dell
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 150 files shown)


src/mgmt/director.cfg
```diff
@@ -1,4 +1,4 @@
-# (c) 2014-2020 Dell
+# (c) 2014-2021 Dell
```
heydog15, line 1: Why does this file just say Dell and all the others have Dell Inc. or its subsidiaries?

src/mgmt/osp-sah.ks
```diff
@@ -1,4 +1,4 @@
-# (c) 2014-2020 Dell
+# (c) 2014-2021 Dell
```
heydog15, line 1: Why does this file just say Dell and all the others have Dell Inc. or its subsidiaries?

CalvinSuma: That is what it said from the start. There were multiple version did you want them all standardized?

heydog15:
I don't know.  Geal, shouldn't they all be the same?  These are the only 2 out of 150

CalvinSuma:
> I don't know. Geal, shouldn't they all be the same? These are the only 2 out of 150

CalvinSuma:
@heydog15 wrong button there should be a few others that have slightly different messages

gaelrehault:
yes, header format should be the same on all files. possible we missed some last year or that some were added incorrectly during the year.

heydog15:
Gael/David, should they all be the same?  I would think so.

heydog15:
Sorry, thanks

heydog15:
Looks good!!

PR #655, opened 2020-12-16
Updated powerflex templates for all the Core NIC Configurations

usama2490:
- Powerflex templates for all the core NIC configurations have been updated
- Used only a single folder instead of creating another folder for powerflex configuration to make things simple (i.e avoided **4_port_powerflex** and used the existing one "**4_port**")
- We have 2 additional files in all the nic folders the one is **nic_environment_powerflex.yaml** file which will be used in ini file for enabling the powerflex backend for storage and the other one is **powerflexstorage.yaml** which will configure the powerflex nodes.

For example, to enable powerflex, there would be a small change need to make in the file path, like that in INI..
"**nic_env_file=4_port/nic_environment_powerflex.yaml**"

@gaelrehault kindly review the PR 

Regards,


gaelrehault:
Could you please update the templates to use the rhel8 interface names ?

usama2490:
> Could you please update the templates to use the rhel8 interface names ?

Yeah I can do that gael. But interface names are overwritten in these templates with the values defined in the INI file while the deployment.  Existing templates have also the older names for interfaces, do you want me to update them also for the record or just the one defined in the new flex files?

gaelrehault:
Let s update all of them please .

usama2490:
Yeah sure gael, i am on it.

usama2490:
I have updated the interface names and did some additional changes in validation checks in the case of xsp and csp profiles. We have successfully deployed powerflex with **5_port/nic_environment_powerflex.yaml** on R87, so configurations for all the other nfv features should be good. 

PR is good to go.

PR #656, opened 2020-12-16
updated undercloud patches for z3

gaelrehault:


gaelrehault:
removed a couple of patches that landed in z3

heydog15:
Gael successfully deployed this on z3 so merging.

PR #661, opened 2021-01-07
Merge edge PRs

dpaterson:


src/pilot/idrac_card_groupid.patch
```diff
@@ -0,0 +1,16 @@
+--- idrac_card_train.py	2021-01-12 11:09:12.702610525 -0600
++++ idrac_card.py	2021-01-12 11:05:14.941262715 -0600
+@@ -61,7 +61,7 @@
+         fqdd = utils.get_wsman_resource_attr(
+             idrac_attr_xml, namespace, 'FQDD')
+         group_id = utils.get_wsman_resource_attr(
+-            idrac_attr_xml, namespace, 'GroupID')
++            idrac_attr_xml, namespace, 'GroupID', nullable=True)
```
rpioso, line 8: Has this been tested by having `config_idrac.py` reset the iDRAC settings?

src/pilot/install-director.sh
```diff
@@ -361,14 +361,6 @@ for container in "ironic_pxe_tftp"  "ironic_conductor" "ironic_api" ;
...
     echo "## Done"
 
-    echo
-    echo "## Patching Ironic iDRAC driver idrac_card.py on ${container}..."
-    upload_file_to_container "${container}" "${HOME}/pilot/idrac_card_groupid.patch" "/tmp/idrac_card_groupid.patch"
-    run_on_container "${container}" "patch -b -s /usr/lib/python3.6/site-packages/dracclient/resources/idrac_card.py /tmp/idrac_card_groupid.patch"
-    run_on_container "${container}" "rm -f /usr/lib/python3.6/site-packages/dracclient/resources/idrac_card.pyc"
-    run_on_container "${container}" "rm -f /usr/lib/python3.6/site-packages/dracclient/resources/idrac_card.pyo"
-    echo "## Done"
-
```
rpioso: For consistency and sanity, I suggest having it be consistent across all installations, at least on the undercloud.

rpioso (changes requested):
Has this been tested by having `config_idrac.py` reset the iDRAC settings?

src/pilot/install-director.sh
gaelrehault: The versions of the module(s) installed on the undercloud base OS ( and used by config idrac ) are different than the ones running on the undercloud containers. If we wanted to patch both we d need x2 set of patches (double the headache for each zStream update).
In this particular case, since it's only affecting a config idrac use case,  i think patching on the os only is fine.

PR #663, opened 2021-01-12
Adding sanity checks (File requirement + versions) based upon the profile

jproque-dell:


gaelrehault:
Target version should be 3.5.x

heydog15:
Why are we checking for 3.6?  Thought we were using 3.5.1?  Is this PR meant to be in the current release that is in 2 weeks because it is kind of late in the game

jproque-dell:
> Why are we checking for 3.6? Thought we were using 3.5.1? Is this PR meant to be in the current release that is in 2 weeks because it is kind of late in the game

My fault! Will fix it. I confirm we use 3.5.1 version for first drop.

jproque-dell:
Versions fixed. I have also reordered the checks so that the presence check is done before validating the profile.

heydog15 (approved):

PR #668, opened 2021-01-25
only apply patch to powerflex ansible if applicable deployment`

gaelrehault:


jproque-dell (approved):
Good catch! I didn't even think about it, but you are abolutely right, i'll fail if powerflex is not enabled since the RDO package won't be installed.

PR #670, opened 2021-02-02
Merging JS-16.1.20-11 back to main

gaelrehault:


arkadykanevsky:
Why such a huge patch?

gaelrehault:
Over the last year we dropped master and worked of branches because we had several releases going on in parallel 
Not the case for the next release , whole team will work on same version, so we 're pulling all the latest release changes back into master to work off

heydog15:
Merging to get everything to master.

PR #671, opened 2021-02-08
AMD/R7515

gaelrehault:


heydog15:
Do we want to add the 6515's as well since the Ironic team has used them?  Might be good to have the model added for future use.

heydog15:
Looks good.

PR #672, opened 2021-02-11
fixing timezone on overcloud nodes

jproque-dell:
During the deployment, timezone is not set on the overcloud nodes which can lead to inconsistency especially when it comes to ssl certificates.
This patch is for enabling the appropriate timezone which is set by default to America/Chicago but can be overwritten with a parameter within the .ini file.


heydog15:
Looks good.  Make sure it goes to master branch as well.

PR #690, opened 2021-07-16
OCP support.

gaelrehault:
This PR adds initial support for OpenShift on BareMetal deployments.


jproque-dell (approved):
Looks good to me. Merging

PR #693, opened 2021-08-23
Fix handling systems without RAID controllers

kamlesh6808:
Systems without RAID controllers were not detected correctly and proceeded trying to configure RAID.
It proceeded until it correctly determined that there are no RAID controllers but emitted a number of confusing log messages before that.

With this change the check at the very beginning of RAID configuration for no RAID controllers is fixed by also checking for empty lists, so it exits at the beginning of RAID configuration without producing confusing log messages.


src/pilot/assign_role.py
```diff
@@ -745,7 +745,7 @@ def configure_raid(ironic_client, node_uuid, super_role, os_volume_size_gb,
     when RAID configuration failed and return False. Further testing
     should uncover interesting error conditions.'''
 
-    if get_raid_controller_ids(drac_client) is None:
+    if not get_raid_controller_ids(drac_client):
```
ajya, line 748: so this change was necessary because `get_raid_controller_ids` returned empty list `[]` and not `None`?

ajya:
Thanks for the patch!
Couple of comments: 
1) The PR title is too generic. Please update the title to be specific - what is fixed here?
2) Don't know much about JetPack workflow, but is it OK to merge this to master? Previously versioned branches were used.
3) The commit message is written so that it gives impression that there was no handling for no controllers. Looking at the change I can see that some handling was done, but didn't cover all cases(?). Or it never returned None? It returned None sometimes? Please update commit message to be more specific what was fixed and what was broken. Currently, to me it looks that it fixes handling empty controllers (not that it _adds_ it if you get what I mean)

ajya (changes requested):
Thanks for the patch!
Couple of comments:

1. The PR title is too generic. Please update the title to be specific - what is fixed here?
2. Don't know much about JetPack workflow, but is it OK to merge this to master? Previously versioned branches were used.
3. The commit message is written so that it gives impression that there was no handling for no controllers. Looking at the change I can see that some handling was done, but didn't cover all cases(?). Or it never returned None? It returned None sometimes? Please update commit message to be more specific what was fixed and what was broken. Currently, to me it looks that it fixes handling empty controllers (not that it _adds_ it if you get what I mean)

src/pilot/assign_role.py
rpioso: @ajya, yes, that is why. The situation was encountered while testing against a systems that has no RAID controller.

When the current code finds that condition, instead of immediately terminating the configuration of RAID processing, it proceeds along until it finds the condition again. A number of confusing log messages are emitted along the way.

rpioso:
@gaelrehault and I chatted. This PR may be proposed against master, because it solves something and does not introduce a regression. It is very, very low risk.

ajya (approved):
thanks for the updates.

PR #694, opened 2021-08-23
Ces 12338

gaelrehault:
* Updated assign role to handle BOSS-2 disks enumeration 
* disabled overcloud image customization until workaround is in place for nested virtualization not supported with IceLake/Rhel 8.2


heydog15:
Looks good.

PR #696, opened 2021-08-24
Virt-customize changes for IceLake on RHEL 8.2

usama2490:
- This PR covers the alternative workaround for **Virt-customize** issue we are facing on **IceLake**, all under the same **"install-director.sh"** file
- Changes consists of SCP the overcloud image first to SAH node, customized it there and reupload the image to director node, covers powerflex nodes as well
- Tested this change with multiple deployments on **R178** (IceLake) Stamp

@gaelrehault kindly review it and let me know if you have any comments/concerns? Thanks

### DIFF
--- a/src/deploy/osp_deployer/director.py
+++ b/src/deploy/osp_deployer/director.py
@@ -189,8 +189,14 @@ class Director(InfraHost):
                   " --director_ip " + \
                   self.ip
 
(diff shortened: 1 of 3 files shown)


gaelrehault (changes requested):
Looks good, however if you re setting the overcloud image password AND using powerflex, you're copying, editing & re uploading the image twice here; would be more efficient to do one trip.

PR #698, opened 2021-09-17
CES-12484 Patch support for H755 RAID controller

gaelrehault:
Validated on r57

### DIFF
--- a/src/pilot/dracclient_raid.patch
+++ b/src/pilot/dracclient_raid.patch
@@ -1,0 +1,54 @@
+--- raid.py	2021-09-16 14:39:49.035032177 -0500
++++ raid.new	2021-09-16 14:38:50.658107416 -0500
+@@ -911,6 +911,15 @@
+         RAID to JBOD or vice versa.  It does this by only converting the
+         disks that are not already in the correct state.
+ 
++        When converting H755 RAID controller physical disks to non-RAID mode,
++        RAID-0 virtual disks get created for each physical disk and disks moved
++        to 'Online' state.
++
++        This is different from other controllers supporting non-RAID conversion
++        and takes up physical disks that cannot be used for user intended RAID
++        configuration later. H755 RAID controllers are excluded from converting
++        to non-RAID mode leaving disks in 'Ready' state.
++
+         :param mode: constants.RaidStatus enumeration that indicates the mode
+                      to change the disks to.
+         :param controllers_to_physical_disk_ids: Dictionary of controllers and
+@@ -939,7 +948,9 @@
+         physical_disks = self.list_physical_disks()
+ 
+         raid = constants.RaidStatus.raid
++        jbod = constants.RaidStatus.jbod
+ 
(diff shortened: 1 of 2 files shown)


src/pilot/dracclient_raid.patch
```diff
@@ -0,0 +1,54 @@
...
++        jbod = constants.RaidStatus.jbod
+ 
++        all_controllers = self.list_raid_controllers()
+         if not controllers_to_physical_disk_ids:
+             controllers_to_physical_disk_ids = collections.defaultdict(list)
+ 
+@@ -952,6 +963,25 @@
+                         physical_d.controller]
+ 
+                     physical_disk_ids.append(physical_d.id)
```
rpioso, line 32: Please add a blank line after this line to align with the upstream change.

src/pilot/dracclient_raid.patch
```diff
@@ -0,0 +1,54 @@
...
++        # Filter out PERC H755 controller as it creates RAID0 virtual disks
++        # when in non-RAID mode
++        controllers_to_results = {}
++
++
++        if mode == jbod:
```
rpioso, line 38: This chunk is being updated upstream during the code review process. Please change this to the following:

    + +
    + +        controllers_to_results = {}
    + +
    + +        # Filter out PERC H755 controller as it creates RAID-0 virtual disks
    + +        # when in non-RAID mode. Returns conversion result dictionary
    + +        # containing is_commit_required and is_reboot_required key with
    + +        # the value always set to False.
    + +        if mode == jbod:

src/pilot/dracclient_raid.patch
```diff
@@ -0,0 +1,54 @@
...
+         disks that are not already in the correct state.
+ 
++        When converting H755 RAID controller physical disks to non-RAID mode,
++        RAID-0 virtual disks get created for each physical disk and disks moved
++        to 'Online' state.
++
++        This is different from other controllers supporting non-RAID conversion
++        and takes up physical disks that cannot be used for user intended RAID
++        configuration later. H755 RAID controllers are excluded from converting
++        to non-RAID mode leaving disks in 'Ready' state.
```
rpioso, line 14: This comment block has been modified upstream during the code review process. Please change this to the following:

    + +        When converting Dell EMC PERC H755 RAID controller physical disks
    + +        to non-RAID mode, RAID-0 virtual disks get created for each physical
    + +        disk and disks moved to 'Online' state.
    + +
    + +        This is different from other controllers supporting non-RAID conversion
    + +        and takes up physical disks that cannot be later used for user intended
    + +        RAID configuration. H755 RAID controllers are excluded when converting
    + +        to non-RAID mode leaving disks in 'Ready' state.

rpioso (commented):
The proposed upstream change has been modified during its code review. While those changes do not affect the logic in the production code, I recommend you more closely align this patch with it. That could make it easier to remove this when it is no longer needed.

